PR #21, opened 2025-09-26
feat: Add new Select alpha components

maximo-macchi-cb:
## What changed? Why?
This PR introduces a new `<Select>` component along with many default children components to match Coinbase design's styles. As well, there are new features to various CDS components and a new hook created. 

The new `<Select>` will be released as an alpha component. Our current `<Select>` doesn't support `null` values and we can't introduce this without creating a breaking change due to Typescript. 

New Select features include:

- Multi-select - select more than one option at a time
- Select all - select all or deselect all options at once
- Null option - support for a null option value to reset the Select into an unselected state
- Custom Select trigger component
- Custom Select option component
- No options empty state with ability to use a custom component for empty state
- `style` and `className` props for custom styling
- Improved performance and accessibility

New CDS features include:

- New hook: `useClickOutside`. Calls a callback function when an area outside a specified ref is clicked
- New util method: `findClosestNonDisabledNode`. Finds the closest non-disabled node when given a collection of nodes
- `<InputChip>` now has an `inverseColorScheme` prop to control the inverse color style
- `<InputChip>` uses children versus props to render its content
- `<FocusTrap>` has a new prop called `includeTriggerInFocusTrap` which includes the trigger that renders a focus trap in the keyboard focus order

## ~UI changes~

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/web/src/alpha/select/DefaultSelectControl.tsx
+++ b/packages/web/src/alpha/select/DefaultSelectControl.tsx
@@ -296,6 +296,7 @@ const DefaultSelectControlComponent = memo(
           </Pressable>
         ),
         [
+          accessibilityLabel,
           ariaHaspopup,
           interactableBlendStyles,
           classNames?.controlInputNode,


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ None for this change
</td>
<td></td>
</tr>
</table>

packages/web/src/alpha/select/DefaultSelectControl.tsx
```diff
@@ -0,0 +1,280 @@
...
+import { Chip } from '../../chips/Chip';
+import { InputChip } from '../../chips/InputChip';
+import { HelperText } from '../../controls/HelperText';
+import { InputLabel } from '../../controls/InputLabel';
+import { InputStack } from '../../controls/InputStack';
+import { HStack } from '../../layout/HStack';
+import { AnimatedCaret } from '../../motion/AnimatedCaret';
+import { Pressable } from '../../system/Pressable';
+import { Text } from '../../typography/Text';
+import { findClosestNonDisabledNodeIndex } from '../../utils/findClosestNoneDisabledNodeIndex';
```
adrienzheng-cb: nit: function `findClosestNonDisabledNodeIndex` has a different name from the file name `findClosestNoneDisabledNodeIndex`

packages/mobile/src/alpha/select/DefaultSelectControl.tsx
```diff
@@ -0,0 +1,279 @@
...
+      inputNode={inputNode}
+      labelNode={shouldShowCompactLabel ? null : labelNode}
+      labelVariant={labelVariant}
+      variant={variant}
+      {...props}
+    />
+  );
+};
+
+export const DefaultSelectControl = memo(forwardRef(DefaultSelectControlComponent)) as <
```
haoruikun-cb: This is nit, but should we wrap memo and forwardRef directly at the component like you did in `packages/web/src/alpha/select/DefaultSelectEmptyDropdownContents.tsx`? I see some place you wraps memo and forwardRef directly at the component but sometimes you export them separately like what you did here. It would be nice to align it. My forwardRef and memo are wrapped directly at the component in `RollingNumber`,

packages/mobile/src/alpha/select/Select.tsx
```diff
@@ -0,0 +1,491 @@
...
+    (typeof openProp !== 'undefined' && typeof setOpenProp === 'undefined')
+  )
+    throw Error(
+      'Select component must be fully controlled or uncontrolled: "open" and "setOpen" props must be provided together or not at all',
+    );
+
+  const controlRef = useRef<any>(null);
+
+  useImperativeHandle(ref, () =>
+    Object.assign(controlRef.current, {
```
haoruikun-cb: I haven't tried doing an `Object.assign` inside the `useImperativeHandle` before. Not sure if this is a best practice, but I have tested passing a ref myself, and at least the typescript doesn't complain. As long as you have tested using this ref to do things like `setOpen` or modify the react node it should be fine.

haoruikun-cb: I asked cursor GPT5 if this is a good approach, and this is what cursor said:
image
image


I'm pasting in the code snippet here:
Do this:
```
type SelectHandle = {
  open: () => void;
  setOpen: (v: boolean) => void;
  refs: { reference: React.RefObject<typeof controlRef.current>; floating: null };
  // optionally: element?: typeof controlRef.current;
};

useImperativeHandle(
  ref,
  () => ({
    open,
    setOpen,
    refs: { reference: controlRef, floating: null },
    // element: controlRef.current, // if you need to expose it
  }),
  [open, setOpen],
);
```
If you want the forwarded ref to be the same as controlRef, merge refs instead of useImperativeHandle:
```
// where mergeRefs = (...refs) => (node) => refs.forEach(r => r && (typeof r === 'function' ? r(node) : (r.current = node)));
ref={mergeRefs(ref, controlRef)}
```

I haven't tried this approach recommended by GPT5, if you want to adopt this approach make sure you have tested it.

packages/mobile/src/alpha/select/DefaultSelectOption.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const handlePress = useCallback(() => onPress?.(value), [onPress, value]);
+
+  const multiSelectCheckedAccessibilityState = useMemo(() => {
+    if (selected) return true;
+    if (indeterminate) return 'mixed';
+    return false;
+  }, [selected, indeterminate]);
+
+  // Note: Cell component doesn't support ref forwarding yet, so we can't pass the ref
```
haoruikun-cb, line 77: Good catch. Maybe we can add a ticket in the backlog to add ref forwarding to Cell component. It should be a non breaking change since it just add an extra optional ref prop.

haoruikun-cb: nit: if we are not adding the Cell ref forwarding support with this Select release, maybe we can make a note in the doc saying ref is currently not supported for this sub component and will be supported in future update

packages/common/src/select/useMultiSelect.ts
```diff
@@ -0,0 +1,87 @@
...
+  initialValue: string[] | null;
+};
+
+/**
+ * API returned by the useMultiSelect hook for managing multi-select state
+ */
+export type MultiSelectApi<T extends string> = {
+  /** Current array of selected values */
+  value: T[];
+  /** Handler for toggling selection of one or more values */
```
hcopp: nit: could we clarify here that when it is an array we simply add if not present but individual strings will be toggled?

packages/mobile/src/alpha/select/DefaultSelectControl.tsx
maximo-macchi-cb: This is done because the `memo()` and `forwardRef()` methods don't preserve the generic type parameters. So it needs to be wrapped and have its type asserted with the `as` call. [This SO post](https://stackoverflow.com/questions/58469229/react-with-typescript-generics-while-using-react-forwardref) gives more details

maximo-macchi-cb: I refactored this a bit. Still need to use the type assertion but my refactor is probably a cleaner approach. Lmk what you think!

haoruikun-cb: Yeah I think this is fine

packages/mobile/src/chips/InputChip.tsx
```diff
@@ -29,9 +30,10 @@ export const InputChip = memo(
             testID={testID ? `${testID}-close-icon` : 'input-chip-close-icon'}
           />
         }
+        invertColorScheme={invertColorScheme}
         {...props}
       >
-        {value}
+        {label}
```
adrienzheng-cb: why can't we just use the `children` prop here, so we unified the api between chips and input chips

apps/docs/docs/components/inputs/Select/_webExamples.mdx
```diff
@@ -8,11 +8,11 @@ function DefaultSelect() {
   const options = ['Option 1', 'Option 2', 'Option 3', 'Option 4', 'Option 5', 'Option 6'];
   return (
     <VStack padding={2}>
-      <Select value={value} placeholder="Choose something..." onChange={setValue}>
+      <OldSelect value={value} placeholder="Choose something..." onChange={setValue}>
```
hcopp, line 11: nit: should we keep this as Select and call the other SelectAlpha since we use that term for most references already?

apps/docs/docs/components/inputs/Select/_mobileExamples.mdx
```diff
@@ -18,7 +18,7 @@ const SelectMobile = () => {
 
   return (
     <>
-      <Select value={value} onPress={handleOpenTray} />
+      <OldSelect value={value} onPress={handleOpenTray} />
```
hcopp, line 21: We don't need to make this change on the mobile examples right?

packages/common/src/select/useMultiSelect.ts
hcopp: nit: could we clarify here that when it is an array we simply add if not present but individual strings will be toggled?

packages/mobile/src/chips/InputChip.tsx
maximo-macchi-cb: Good callout! I think this would be better addressed in its own PR. 

We could add `children` support but removing the `label` prop would be a breaking change. It'd be better to break this in v9 since that's most likely when we'd fully remove the `value` prop, which is now marked as deprecated

apps/docs/docs/components/inputs/Select/_webExamples.mdx
maximo-macchi-cb: This is purposely being done to make it a bit more clear to customers which `<Select>` (the new alpha one) should be used. Since customers may copy the code examples, having the newer examples use `<Select>` in its markup will be a better UX

apps/docs/docs/components/inputs/Select/_mobileExamples.mdx
maximo-macchi-cb: Same point as [this comment](https://github.com/coinbase/cds/pull/21#discussion_r2448737335)! ☝️

haoruikun-cb (approved):

PR #28, opened 2025-09-27
Get rid of nullish check & Scrubber context refactor

cb-ekuersch:
## What changed? Why?

This PR does 2 major this:
1. remove unnecessary nullish check on getScale/Axis function from ChartContext
2. refactors the ScrubberContextProvider into its own component which encapsulates all the scrubbing event handling logic

This second part achieves 2 things:
1. Massively cleans up Chart.tsx
2. Enforces better separation of concerns as now the ScrubberProvider component encapsulates 100% of all scrubber related state and callbacks

This setup came at a small cost:
The ScrubberProvider requires a ref to the root SVG element in order to attach event handlers. This meant that we could no longer reuse the full Context Type in common. To get around this I made the ContextValue type generic to pass in the type of the ref element and then created the contexts in each platform with their own hooks. This can be found in scrubber/ScrubberProvider.tsx

## UI changes

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


packages/common/src/visualizations/charts/context.ts
```diff
@@ -6,7 +6,11 @@ import type { AxisConfig } from './axis';
 import type { Series } from './chart';
 import type { ChartScaleFunction } from './scale';
 
-export type ChartContextValue = {
+export type ChartContextValue<TRef> = {
```
cb-ekuersch, line 9: The big change to common is here + remove the ChartContext and hook from this package. Those are now set up in each package

PR #31, opened 2025-09-29
feat: list cell

haoruikun-cb:
## What changed? Why?
Added `layoutSpacing` prop. If `layoutSpacing='hug'`, it will use our new ListCell design.

### New ListCell design ([figma](https://www.figma.com/design/k5CtyJccNQUGMI5bI4lJ2g/branch/cVoQb7emq3cv2bnpcQSmOf/%E2%9C%A8-CDS-Components?node-id=155-12186&p=f&m=dev)): 
1. padding adjustment
2. Always hug content, no minHeight predefined for customers
3. Title cap at 2 lines always
4. 0 border radius for pressable
6. no outer padding for pressable
7. Description & Subdetail font change to `label2` 

### `Cell` changes:
1. Added `classNames` and `styles` props.
2. Deprecated `detail` and replace with `end`

### `ListCell` changes:
1. Added `classNames` and `styles` props.
3. Added `layoutSpacing` props.
4. Changed description to `label2` when `layoutSpacing='hug'`
5. Deprecated `action` and replace with `end`

### `CellDetail` changes:
1. Change sub detail font to `label2` when `layoutSpacing='hug'`

### `CellMedia` changes:
1. Mark this component as deprecated. We move away from using `CellMedia` component for media but ask users to directly pass in media like icon, avatar, etc.

## UI changes

| image |image|

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web
- [x] Manual - Android (Emulator / Device)

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.16.0 ((10/8/2025, 01:28 PM PST))
+
+This is an artificial version bump with no new change.
+
 ## 8.15.0 ((10/8/2025, 6:30 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-common",
-  "version": "8.15.0",
+  "version": "8.16.0",
   "description": "Coinbase Design System - Common",
   "repository": {
     "type": "git",
--- a/packages/mcp-server/CHANGELOG.md
+++ b/packages/mcp-server/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.16.0 ((10/8/2025, 01:28 PM PST))
+
+This is an artificial version bump with no new change.
+
 ## 8.15.0 ((10/8/2025, 6:30 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/mcp-server/package.json
+++ b/packages/mcp-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-mcp-server",
-  "version": "8.15.0",
+  "version": "8.16.0",
   "description": "Coinbase Design System - MCP Server",
   "repository": {
     "type": "git",
--- a/packages/mobile/CHANGELOG.md
+++ b/packages/mobile/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.16.0 (10/8/2025 PST)
+
+#### 🚀 Updates
+
+- New ListCell in layoutSpacing variants. [[#31](https://github.com/coinbase/cds/pull/31)]
+
 ## 8.15.0 ((10/8/2025, 6:30 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-mobile",
-  "version": "8.15.0",
+  "version": "8.16.0",
   "description": "Coinbase Design System - Mobile",
   "repository": {
     "type": "git",
--- a/packages/web/CHANGELOG.md
+++ b/packages/web/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.16.0 (10/8/2025 PST)
+
+#### 🚀 Updates
+
+- New ListCell in layoutSpacing variants. [[#31](https://github.com/coinbase/cds/pull/31)]
+
 ## 8.15.0 (10/8/2025 PST)
 
 ### 🚀 Updates
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-web",
-  "version": "8.15.0",
+  "version": "8.16.0",
   "description": "Coinbase Design System - Web",
   "repository": {
     "type": "git",


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ None for this change
</td>
<td></td>
</tr>
</table>

packages/web/src/alpha/cells/CellDetail.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const variantColorMap: Record<CellDetailVariant, ThemeVars.Color> = {
+  foregroundMuted: 'fgMuted',
+  negative: 'fgNegative',
+  positive: 'fgPositive',
+  warning: 'fgWarning',
+};
+
+export const CellDetail = memo(function CellDetail({
```
cb-ekuersch: Is this component even necessary?

packages/web/src/alpha/cells/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ListCell';
```
cb-ekuersch: `export * from './Cell';`

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+  }
+  &:focus-visible {
+    outline-style: solid;
+    outline-width: 2px;
+    outline-color: var(--color-bgPrimary);
+    outline-offset: 0;
+  }
+`;
+
+const baseCss = css`
```
cb-ekuersch: nit: Box component should already be display:block

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+`;
+
+export const cellDefaultElement = 'div';
+
+export type CellDefaultElement = typeof cellDefaultElement;
+
+export type CellBaseProps = Polymorphic.ExtendableProps<
+  BoxBaseProps,
+  Pick<PressableProps<'a'>, 'href' | 'target'> & {
+    onKeyDown?: React.KeyboardEventHandler;
```
cb-ekuersch: does Box not already have a lot of these props?

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+    onKeyUp?: React.KeyboardEventHandler;
+    onClick?: React.MouseEventHandler;
+    accessory?: React.ReactElement<CellAccessoryProps>;
+    children: React.ReactNode;
```
cb-ekuersch: we should consider dropping children in this new API. Imo it's difficult to reason about where this slots in with the other slots

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+    end?: React.ReactNode;
+    intermediary?: React.ReactNode;
+    start?: React.ReactElement;
+    // TODO: consider renaming this to shouldTruncate. Since overflow gives people the sense that it will overflow and overlap with other content
```
cb-ekuersch: +1

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+    shouldOverflow?: boolean;
+    borderRadius?: ThemeVars.BorderRadius;
```
cb-ekuersch: Box has this

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+    // TODO: consider removing this since we have styles.end
```
cb-ekuersch: +1

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+    /** Apply a fixed width to the end (end). */
+    endWidth?: number | string;
+    /** Is the cell disabled? Will apply opacity and disable interaction. */
+    disabled?: boolean;
+    /** Which piece of content has the highest priority in regards to text truncation, growing, and shrinking. */
+    priority?: CellPriority | CellPriority[];
+    /** Is the cell selected? Will apply a background and selected accessory. */
+    selected?: boolean;
```
cb-ekuersch: should this be moved up to ListCell? This feels like kinda of specific behavior for the generic layout component

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+        classNames,
+        styles,
+        style,
+        ...props
+      }: CellProps<AsComponent>,
+      ref?: Polymorphic.Ref<AsComponent>,
+    ) => {
+      const Component = (as ?? cellDefaultElement) satisfies React.ElementType;
+      const isAnchor = Boolean(href);
+      const isButton = Boolean(onClick ?? onKeyDown ?? onKeyUp);
```
cb-ekuersch: why would onKeyDown/onKeyUp indicate this should behave as a button? Also we should probably add another clause: `!isAnchor` since an anchor can definitely have these event handlers as well

packages/web/src/alpha/cells/Cell.tsx
```diff
@@ -0,0 +1,392 @@
...
+          className: cx(pressCss, insetFocusRingCss, classNames?.pressable),
+          disabled,
+          onClick,
+          onKeyDown,
+          onKeyUp,
+          tabIndex,
+          testID: testID && `${testID}-cell-pressable`,
+          style: styles?.pressable,
+        };
+        if (isAnchor)
```
cb-ekuersch: maybe:
```
if (href || onClick) {
  return (
    <Pressable as={href ? "a" : "button"} href={href} target={target} {...pressableSharedProps}>
        {content}
     </Pressable>
  )
}
```

packages/web/src/alpha/cells/Cell.tsx
haoruikun-cb: Now if we are going with the 'layoutDensity' variant approach, probably we need to avoid changing props name for now. Potentially leave it to future breaking change.

packages/web/src/cells/Cell.tsx
```diff
@@ -84,8 +84,9 @@ export type CellBaseProps = Polymorphic.ExtendableProps<
     detail?: React.ReactNode;
     intermediary?: React.ReactNode;
     media?: React.ReactElement;
+    // TODO: consider renaming this to shouldTruncate in next breaking change release. Since overflow gives people the sense that it will overflow and overlap with other content
     shouldOverflow?: boolean;
-    borderRadius?: ThemeVars.BorderRadius;
+    // TODO: consider removing this in next breaking change release. Since we have styles.detail
```
cb-ekuersch: let's @deprecate it then and point them to styles.detail

packages/web/src/cells/Cell.tsx
```diff
@@ -100,6 +101,30 @@ export type CellBaseProps = Polymorphic.ExtendableProps<
     innerSpacing?: CellSpacing;
     /** The content to display below the main cell content */
     bottomContent?: React.ReactNode;
+    /** Styles for the components */
+    styles?: {
+      root?: React.CSSProperties;
+      contentContainer?: React.CSSProperties;
```
cb-ekuersch, line 133: I recommend upating the docs for this component to include a section on the styling. These names are very ambiguous from this perspective

packages/web/src/cells/ListCell.tsx
```diff
@@ -21,15 +22,39 @@ export const listCellDefaultElement = 'div';
...
     compact?: boolean;
+    /**
+     * When 'dense' is set, the cell will have the following behavior:
+     * 1. No min-height, height is determined by the content
+     * 2. smaller padding, no extra padding around the pressable area
+     * 3. 0 border radius for pressable shade
+     * 4. Title always cap at 2 lines
+     * 5. Description and subdetail have smaller font
+     */
+    layoutDensity?: 'sparse' | 'compact' | 'dense';
```
cb-ekuersch: isn't all this more about padding instead of "density"? Maybe we can rename the prop to align with the concept. Perhaps `layoutSpacing`? You could use Cursor to help with the name. In this case our new mode would actually be `"none"` which sounds more honest

packages/web/src/cells/ListCell.tsx
```diff
@@ -53,6 +78,40 @@ export type ListCellBaseProps = Polymorphic.ExtendableProps<
     multiline?: boolean;
     /** Title of content. Max 1 line (with description) or 2 lines (without), otherwise will truncate. */
     title?: React.ReactNode;
+    /** Class names for the components */
+    classNames?: {
```
cb-ekuersch: can you `Pick` these from Cell?

apps/docs/docs/components/media/CellMedia/_mobileExamples.mdx
```diff
@@ -1,3 +1,7 @@
+:::warning Deprecated
```
adrienzheng-cb, line 1: nice

apps/docs/docs/components/data-display/ListCell/_webExamples.mdx
```diff
@@ -124,3 +141,70 @@ The ListCellFallback component provides loading state representations of ListCel
...
+```
+
+With helper text (top + bottom layout):
+
+```text
+┌─────────────────────────────────────────────────────────────────────────────┐
+│                                 root (Box)                                  │
+│┌───────────────────────────────────────────────────────────────────────────┐│
+││                                pressable                                  ││
+││┌─────────────────────────────────────────────────────────────────────────┐││
```
adrienzheng-cb, line 213: Do we need the contentContainer layer, or we can just pass `flexDirection` to Pressable

packages/mobile/src/cells/ListCell.tsx
```diff
@@ -30,7 +76,7 @@ export type ListCellBaseProps = CellDetailProps &
      * the title will take up a single line truncating with ellipses.
      */
     disableMultilineTitle?: boolean;
-    /** Assitive message to display below the cell content */
+    /** Assistive message to display below the cell content. */
```
adrienzheng-cb, line 87: nice catch

apps/docs/docs/components/data-display/ListCell/_webExamples.mdx
haoruikun-cb: I also thought about it, but I'm hesitant to remove this internal layer at this point, since this will be a breaking change (especially we already exposed `contentClassName` prop and customers may have already add some style override to it).

packages/mobile/src/cells/Cell.tsx
```diff
@@ -34,13 +34,22 @@ export type CellSpacing = Pick<
...
     detail?: React.ReactNode;
+    /** Middle content between main content and detail. */
     intermediary?: React.ReactNode;
+    /** Media rendered at the start of the cell (icon, avatar, image, etc). */
     media?: React.ReactElement;
     borderRadius?: ThemeVars.BorderRadius;
-    /** Apply a fixed width to the detail (end). */
+    /**
+     * Apply a fixed width to the detail (end).
+     * @deprecated Use `styles.detail.end` instead. This prop is kept for backward
```
cb-ekuersch: isn't it just styles.end? This also got me thinking: the name of the style and the name of the slot (the prop itself) don't match up (i.e. end vs detail). I wonder if we should deprecate detail as a whole and just switch it to `end`. `detail` could be the prop for ListCell since its more specific

haoruikun-cb: That's a very good point. Initially I was also annoyed by this difference but I hesitant to change the prop name since it will be a breaking change. However, I think your deprecation strategy works here. I also went ahead and deprecate `action` in ListCell to call it `end`. Since it's not only limited to actions, it could be anything user customized, so it's more accurate to just call it end: https://github.com/coinbase/cds/pull/31/commits/7163b38bd74441fa325c5dfb09220337bc96f4e4

packages/mobile/src/cells/Cell.tsx
```diff
@@ -34,13 +34,30 @@ export type CellSpacing = Pick<
...
+    /** Main content of the cell; typically title/description content. */
     children: React.ReactNode;
+    /**
+     * End-aligned content (e.g., value, status).
+     * Replaces the deprecated `detail` prop.
+     */
+    end?: React.ReactNode;
+    /**
+     * End-aligned detail content (e.g., value, status).
+     * @deprecated Use `end` instead. Maintained for backward compatibility.
```
hcopp: Could you update this to say

> @deprecated Use `end` instead.

or

> @deprecated Use `end` instead. `detail` will be removed in a future release.

packages/mobile/src/cells/CellDetail.tsx
```diff
@@ -18,7 +18,15 @@ export type CellDetailProps = {
   subdetail?: React.ReactNode;
   /** Variant color to apply to the subdetail text. */
   variant?: CellDetailVariant;
+  /** Specifies whether font should be scaled down automatically to fit given style constraints. */
   adjustsFontSizeToFit?: boolean;
+  /** Layout spacing configuration. */
+  layoutSpacing?: 'spacious' | 'compact' | 'hug';
```
hcopp: Is there a default for this? Trying to understand what will happen here if someone doesn't use hug

packages/mobile/src/cells/CellDetail.tsx
```diff
@@ -46,8 +61,9 @@ export const CellDetail = memo(function CellDetail({
         <Text
           adjustsFontSizeToFit={adjustsFontSizeToFit}
           color={variantColorMap[variant]}
-          font="body"
+          font={layoutSpacing === 'hug' ? 'label2' : 'body'}
```
hcopp: Maybe we don't need to share layoutSpacing with this component? And can default to body but if we support some way of overriding that would fix it.

packages/mobile/src/cells/ListCell.tsx
```diff
@@ -1,21 +1,75 @@
...
+  marginX: 0 as const,
+};
+// no padding outside of the pressable area
+export const hugOuterSpacing: CellSpacing = {
+  paddingX: 0 as const,
+  paddingY: 0 as const,
+  marginX: 0 as const,
+};
+
+type CellStyles = NonNullable<CellBaseProps['styles']>;
```
hcopp, line 24: Can you explain the need for non nullable here?

apps/docs/docs/components/data-display/ListCell/_mobileExamples.mdx
```diff
@@ -1,23 +1,29 @@
...
+:::
+
 ### Basic usage
 
 ```tsx
-<ListCell title="Basic List Cell" description="A simple example of ListCell" />
+<ListCell layoutSpacing="hug" title="Basic List Cell" description="A simple example of ListCell" />
 ```
 
 ### With Media
```
hcopp, line 47: Should we have a section here which is called 'layout spacing' or something to showcase the different designs? You could keep the note above or move it there but that way teams will have a better understanding of how it impacts things before they see the JSDocs in the code - and how to update their designs before we drop the prev layout options.

packages/web/src/cells/ListCell.tsx
```diff
@@ -21,15 +22,70 @@ export const listCellDefaultElement = 'div';
...
     accessory?: CellAccessoryType;
-    /** Interactive action, like a CTA or form element. Cannot be used alongside `onPress`. */
+    /**
+     * End-aligned content (e.g., CTA, form element, metric). Replaces the deprecated `action` prop.
+     * If the content is an action (like button, link, etc), we recommend avoiding use alongside `onClick`.
+     */
+    end?: React.ReactNode;
+    /**
+     * Interactive action, like a CTA or form element. Cannot be used alongside `onClick`.
+     * @deprecated Use `end` instead. Maintained for backward compatibility.
```
hcopp: Thanks for adding these JSDocs. Could we specify what happens if onClick is used alongside action?

I don't know what happens off the top of my head but if action is used and that replaces onClick it could say

> * Interactive action, like a CTA or form element. When specified, overrides any `onClick` handler.

Or

> * Interactive action, like a CTA or form element. When `onClick` is specified, that takes priority over `action`.

packages/web/src/cells/ListCell.tsx
```diff
@@ -21,15 +22,70 @@ export const listCellDefaultElement = 'div';
...
+
 export type ListCellBaseProps = Polymorphic.ExtendableProps<
   Omit<CellBaseProps, 'children'>,
   CellDetailProps & {
     /** Accessory to display at the end of the cell. */
     accessory?: CellAccessoryType;
-    /** Interactive action, like a CTA or form element. Cannot be used alongside `onPress`. */
+    /**
+     * End-aligned content (e.g., CTA, form element, metric). Replaces the deprecated `action` prop.
+     * If the content is an action (like button, link, etc), we recommend avoiding use alongside `onClick`.
```
hcopp, line 47: nit: could say `Replacement for the deprecated `action` prop, and takes precedence over it.`

hcopp (commented):
Implementation looks good! Just have discussion points over JSDocs and some types. Most of my comments can be applied in multiple places.

packages/mobile/src/cells/CellDetail.tsx
haoruikun-cb: It will default to 'spacious', which is the old default.

packages/mobile/src/cells/ListCell.tsx
haoruikun-cb: image
need to have nonnullable to make the type declaration here work
In this image is where I remove nonnullable

packages/mobile/src/cells/CellDetail.tsx
haoruikun-cb: Now I feel like keeping the textProps is helpful for this kind of usage:
image

hcopp: Couldn't you do styles={{ subdetail: { ... }} ? You might have to set a few different properties but in v9 we would solve this by not wrapping react nodes in a Text component

haoruikun-cb: The styles will work but it requires overriding a bunch of CSS like fontSize, fontFamily, fontWeight whereas font prop is just 'label2'.
We can allow them to pass in react node that's not being wrapped in a Text component in V9, but when they do pass in string, we need to have a predefined font for them.

cb-heimdall:
Review Error for hcopp @ 2025-10-08 20:34:04 UTC
User failed mfa authentication, either user does not exist or public email is not set on your github profile. `\
        `see [go/mfa-help](http://go/mfa-help)

PR #96, opened 2025-10-10
feat: Improve Tabs components keyboard navigation

maximo-macchi-cb:
## What changed? Why?
This PR improves the keyboard navigation experience for our `<Tabs>` component to more closely align with the WCAG standard. The following additions have been added:
- The `<Tabs>` component has event handlers for the `ArrowLeft` / `ArrowRight` / `Home` / `End` keys to shift focus between tabs
- Pressing `Tab` shifts focus to the next focusable element and out of the tablist. This avoids users needing to tab through multiple tab options
- ARIA roles for the SegmentedTabs components have been updated

Note that no changes to the mobile package have been made since this PR's focus is improving keyboard navigation. 

## ~UI changes~ n/a

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/web/src/tabs/Tabs.tsx
+++ b/packages/web/src/tabs/Tabs.tsx
@@ -46,7 +46,17 @@ export type TabsActiveIndicatorProps = {
 } & BoxProps<BoxDefaultElement> &
   MotionProps;
 
-export type TabComponent<T extends string = string> = React.FC<TabValue<T> & { tabIndex?: number }>;
+export type TabComponentProps<T extends string = string> = TabValue<T> & {
+  /** The tab index for the tab. Automatically set to manage focus behavior. */
+  tabIndex?: number;
+  /**
+   * The role for the tab.
+   * @default "tab"
+   */
+  role?: string;
+};
+
+export type TabComponent<T extends string = string> = React.FC<TabComponentProps<T>>;
 
 export type TabsActiveIndicatorComponent = React.FC<TabsActiveIndicatorProps>;
 
@@ -203,6 +213,7 @@ const TabsComponent = memo(
                     data-rendered-tab
                     disabled={tabDisabled}
                     id={id}
+                    role="tab"
                     tabIndex={activeTab?.id === id || !activeTab ? 0 : -1}
                     {...props}
                   />
--- a/apps/docs/docs/components/navigation/Tabs/_webExamples.mdx
+++ b/apps/docs/docs/components/navigation/Tabs/_webExamples.mdx
@@ -16,7 +16,6 @@
         onClick={() => api.updateActiveTab(id)}
         disabled={disabled}
         aria-pressed={isActive}
-        role="tab"
         {...props}
       >
         <TabLabel id={id} active={isActive}>
@@ -67,7 +66,6 @@
         onClick={() => updateActiveTab(id)}
         disabled={disabled}
         aria-pressed={isActive}
-        role="tab"
         {...props}
       >
         <TabLabel id={id} active={isActive}>


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/tabs/Tabs.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { forwardRef, memo, useCallback, useMemo } from 'react';
+import type { KeyboardEvent } from 'react';
```
codynova: Don't import types from React, use the global React namespace directly instead

packages/web/src/tabs/Tabs.tsx
```diff
@@ -89,11 +90,11 @@ const TabsComponent = memo(
       const api = useTabs<T>({ tabs, activeTab, disabled, onChange });
       const activeTabRef = activeTab ? refMap.getRef(activeTab.id) : null;
 
-      const [tabsContainerRef, tabsContainerRect] = useMeasure({
+      const [tabsContainerRectRef, tabsContainerRect] = useMeasure({
```
codynova: I'm not sure this name change makes sense, `tabsContainerRef` makes it sound like it's a ref for the tabs container element, whereas `tabsContainerRectRef` makes it sound like its a ref for the tabs container rect. The ref is for the element itself, not the rect

packages/web/src/tabs/Tabs.tsx
```diff
@@ -106,6 +107,88 @@ const TabsComponent = memo(
         };
       }, [activeTabRef, tabsContainerRect]);
 
+      const handleTabsContainerKeyDown = useCallback(
+        (e: KeyboardEvent<HTMLElement>) => {
```
codynova: Don't abbreviate event name

packages/web/src/tabs/Tabs.tsx
```diff
@@ -106,6 +107,88 @@ const TabsComponent = memo(
...
+          }
+
+          if (!focusedTabId) return;
+
+          const focusedTabIndex = tabs.findIndex((tab) => tab.id === focusedTabId);
+          if (focusedTabIndex === -1) return;
+
+          let targetIndex: number | null = null;
+
+          switch (e.key) {
```
codynova: I think this would read a lot simpler and cleaner as if/else

packages/web/src/tabs/Tabs.tsx
```diff
@@ -106,6 +107,88 @@ const TabsComponent = memo(
...
+              }
+              break;
+            }
+          }
+
+          if (targetIndex !== null) {
+            const targetTab = tabs[targetIndex];
+            const targetRef = refMap.getRef(targetTab.id);
+            // Focus the first focusable element within the target tab
+            const focusableElement = targetRef?.querySelector<HTMLElement>(
```
codynova, line 169: Is this actually guaranteed to be a button? Should we be using data-attributes?

packages/web/src/tabs/Tabs.tsx
```diff
@@ -106,6 +107,88 @@ const TabsComponent = memo(
...
+
+          switch (e.key) {
+            case 'ArrowRight': {
+              e.preventDefault();
+              for (let i = focusedTabIndex + 1; i < tabs.length; i++) {
+                if (!tabs[i].disabled) {
+                  targetIndex = i;
+                  break;
+                }
+              }
```
codynova: Do we need to do this index based finding? I think maybe this could be simplified

```tsx
let targetTab: TabValue | null = null;
if (event.key === 'ArrowRight') {
  targetTab = tabs.slice(focusedTabIndex + 1).find((tab) => !tab.disabled);
}
// etc.

if (targetTab === null) return;
const tabRef = refMap.getRef(targetTab.id);
targetRef?.querySelector<HTMLElement>('data-tab-control')?.focus();
```

packages/web/src/tabs/Tabs.tsx
maximo-macchi-cb: I'm fine to revert it! Definitely not needed for this PR's changes

packages/web/src/tabs/Tabs.tsx
maximo-macchi-cb: I used the `.find()` method for iterating forward but for `ArrowLeft` and `Home` key events we need to iterate backwards

packages/web/src/tabs/Tabs.tsx
maximo-macchi-cb: Ah, good idea! I added a data attribute to the `<RenderedTab>` component in the markup

packages/web/src/chips/TabbedChips.tsx
```diff
@@ -26,6 +26,7 @@ const TabComponent = <T extends string = string>({ label = '', id, ...tabProps }
       inverted={isActive}
       onClick={handleClick}
       role="tab"
+      tabIndex={isActive || !activeTab ? 0 : -1}
```
hcopp: Can you look at adding this to the new alpha TabbedChips?

packages/web/src/tabs/Tabs.tsx
```diff
@@ -145,7 +199,7 @@ const TabsComponent = memo(
               const RenderedTab = CustomTabComponent ?? TabComponent;
               return (
                 <TabContainer key={id} id={id} registerRef={registerRef}>
-                  <RenderedTab disabled={tabDisabled} id={id} {...props} />
+                  <RenderedTab data-rendered-tab disabled={tabDisabled} id={id} {...props} />
```
hcopp: Is there something we can do to get this working on our Tabs example in the doc site like it works for SegmentedTabs and TabbedChips? http://localhost:3000/components/navigation/Tabs/

maximo-macchi-cb: I spread the example tab component's props so that the `data-rendered-tab` is applied to the correct element. Good catch!

packages/web/src/tabs/Tabs.tsx
```diff
@@ -46,7 +46,7 @@ export type TabsActiveIndicatorProps = {
 } & BoxProps<BoxDefaultElement> &
   MotionProps;
 
-export type TabComponent<T extends string = string> = React.FC<TabValue<T>>;
+export type TabComponent<T extends string = string> = React.FC<TabValue<T> & { tabIndex?: number }>;
```
hcopp: Could you add export type TabComponentProps which includes the tab value and then tabIndex?

And then have TabComponent which has React.FV<TabComponentProps<T>>. And in the JSDocs for tabIndex you can explain it is automatically set or something for a11y.

This way a component being made off of this prop will work better, such as https://github.com/coinbase/cds/blob/master/apps/docs/src/components/page/ComponentTabsContainer/index.tsx#L31

hcopp: Actually in my goal to get the component tab (switching from examples to props) to work I had to pass in role="tab" as well

```
const CustomTab = ({ id, label, ...props }: TabComponentProps) => {
  const { activeTab, updateActiveTab } = useTabsContext();
  const isActive = activeTab?.id === id;
  return (
    <Pressable
      color={isActive ? 'fgPrimary' : 'fg'}
      font="headline"
      onClick={() => updateActiveTab(id)}
      paddingBottom={0.75}
      paddingTop={0.5}
      role="tab"
      {...props}
    >
      {label}
    </Pressable>
  );
};
```

Should we pass that in as well? And have role be a prop they should accept? Or do you think it is fine since we showcase this in doc examples... I do see it being pretty similar to tabIndex as in including one might be better to include both or neither? Or at least clarify this in docs site in addition the examples that are already updated

maximo-macchi-cb: Good point! I added in that `TabComponentProps` type and updated how `role` is set so it has a default value. But customers can still override the `role` attribute in their custom tab component if they'd like

PR #139, opened 2025-10-31
feat: add Calendar component and replace mobile native date picker

sverg-cb:
## What changed? Why?

New Features:

* [Mobile] Standalone Calendar component cloned from cds-web and updated for cds-mobile

Enhancements:

* [Mobile] Comprehensive accessibility with screen reader support for Calendar component
* [Mobile] Replaced react-native-date-picker with Calendar component for DatePicker
* [Mobile + Web] Added/replaced calendar style props to match the new "classnames/styles" APIs
* [Mobile] Added custom a11y labels for Calendar navigation
* [Mobile] Added `accessibilityState` to Tooltip trigger

Dependencies:

* [Mobile] Remove react-native-date-picker (no longer needed)

Tests:

* [Mobile] Add Storybook for Calendar, including DatePicker, standalone, and custom triggers
* [Mobile] Add unit tests for Calendar and DatePicker
* [Web] Add slot styling examples for Calendar storybook

Documentation:

* [Mobile] Added mobile docs for Calendar
* [Mobile + Web] Updated Calendar and DatePicker documentation to reflect the updated style APIs

### Calendar

| N/A | Screenshot_20260309_124735 |

| N/A | IMG_5190 |

### DatePicker

| mdp_android_old | Screenshot_20260309_125309 |

| mdp_ios_old | IMG_5189 |

The above iOS screenshots were taken on a physical iPhone 15 Pro. Below are 2 additional screenshots from iOS devices of screen sizes smaller and larger than that of the 15 Pro.

| iPhone SE (3rd generation)  | iPad Pro (11-inch) |
|--------|--------|
| Simulator Screenshot - iPhone SE (3rd generation) - 2026-03-15 at 21 04 51 | Simulator Screenshot - iPad Pro 11-inch (M4) - 2026-03-12 at 13 19 14 |

### Accessibility

Using the Android Emulator with TalkBack

[Screen_recording_20260309_130525.webm](https://github.com/user-attachments/assets/019acf89-0484-453b-a2d0-122ddeecee54)

Using a physical iPhone with VoiceOver

https://github.com/user-attachments/assets/529780d1-f1f8-4322-9f61-651979dbb282

### Documentation

| Calendar | DatePicker |
| Screenshot 2026-03-09 at 12 58 01 PM | Screenshot 2026-03-09 at 12 57 42 PM |

### How has it been tested?

- [x] Unit tests
- [x] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web
- [x] Manual - Android (Emulator / Device)

### Testing instructions

For unit and interaction tests, use `yarn nx test mobile`.

Refer to the [mobile-app docs](https://github.com/coinbase/cds/blob/master/apps/mobile-app/README.md) for information on how to test via Expo.

View local Docusaurus changes with `yarn nx dev docs`.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/mobile/src/dates/Calendar.tsx
+++ b/packages/mobile/src/dates/Calendar.tsx
@@ -308,7 +308,6 @@ export const Calendar = memo(
         previousArrowAccessibilityLabel = 'Go to previous month',
         todayAccessibilityHint = 'Today',
         highlightedDateAccessibilityHint = 'Highlighted',
-        testID,
         style,
         styles,
         ...props


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/mobile/src/overlays/tooltip/TooltipProps.ts
```diff
@@ -66,6 +72,13 @@ export type TooltipBaseProps = SharedProps &
      * is read correctly for voice-overs.
      */
     accessibilityHint?: SharedAccessibilityProps['accessibilityHint'];
+    /**
+     * Indicates that the trigger element represents disabled content.
+     * When true, screen readers will perceive the element as disabled,
+     * but the tooltip remains tappable for sighted users.
+     * @default false
```
hcopp: nit: no need to specify a default of false

packages/mobile/src/overlays/tooltip/Tooltip.tsx
```diff
@@ -24,6 +25,7 @@ export const Tooltip = memo(
     visible,
     invertColorScheme = true,
     elevation,
+    triggerDisabled = false,
```
hcopp:
```suggestion
    triggerDisabled,
```

apps/mobile-app/src/routes.ts
```diff
@@ -121,6 +121,10 @@ export const routes = [
     getComponent: () =>
       require('@coinbase/cds-mobile/buttons/__stories__/ButtonGroup.stories').default,
   },
+  {
+    key: 'Calendar',
+    getComponent: () => require('@coinbase/cds-mobile/dates/__stories__/Calendar.stories').default,
+  },
```
hcopp, line 127: One annoying thing, make sure to run `yarn nx run codegen:mobile-routes` when you add new routes. There are multiple spots where it gets added.

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -82,12 +119,17 @@ export const DatePicker = memo(
         invalidDateError = 'Please enter a valid date',
         disabledDateError = 'Date unavailable',
         label,
+        accessibilityHint = 'Enter date or select from calendar using the calendar button.',
         accessibilityLabel,
         accessibilityLabelledBy,
-        calendarIconButtonAccessibilityLabel,
+        calendarIconButtonAccessibilityLabel = 'Open calendar',
+        handleBarAccessibilityLabel = 'Close calendar',
```
hcopp: I want to think about this more. I know on web we just have `calendarIconButtonAccessibilityLabel`, but since it is not a controlled state someone would need to listen to onOpen and onClose to update the label.

Vs for Carousel with play/pause we support a playAccessibilityLabel and pauseAccessibilityLabel...

So in this case it could be [open/close]AccessibilityLabel props

And then we could simply deprecate the old props/fallback to them if the new one isn't set

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,518 @@
...
+  /** Custom style to apply to the Calendar container. */
+  style?: StyleProp<ViewStyle>;
+};
+
+export type CalendarRefHandle = {
+  /** Sets accessibility focus on the selected date, seed date, or today. */
+  focusInitialDate: () => void;
+};
+
+export type CalendarProps = CalendarBaseProps & Omit<VStackProps, 'children' | 'ref'>;
```
hcopp: Could be a great chance to add styles props to mobile Calendar and styles/classnames to web Calendar

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -39,20 +41,49 @@ export type DatePickerProps = {
...
   /**
    * Accessibility label describing the calendar IconButton, which opens the calendar when pressed.
-   * @default 'Open calendar' / 'Close calendar'
+   * @default 'Open calendar'
+   * @deprecated Use openCalendarAccessibilityLabel instead
+   * @note This prop is deprecated and will be removed in a future version.
    */
   calendarIconButtonAccessibilityLabel?: string;
+  /**
+   * Accessibility label for the calendar IconButton, which opens the calendar when pressed.
```
hcopp, line 57: nit this should say ` @default 'Open calendar'` and then above drop the default prop and say`@deprecated Use openCalendarAccessibilityLabel/closeCalendarAccessibilityLabel instead`

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -39,20 +41,49 @@ export type DatePickerProps = {
...
+   */
+  confirmButtonAccessibilityLabel?: string;
+  /**
+   * Accessibility hint for the confirm button in its disabled state.
+   * Only applies when no date is selected. When a date is selected, no hint is shown.
+   * @default 'Select a date first'
+   */
+  confirmButtonDisabledAccessibilityHint?: string;
+  /** Custom style to apply to the Calendar container. */
+  calendarStyle?: StyleProp<ViewStyle>;
```
hcopp: You can look at styles props from other components like Select or Carousel to see how we handle those

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -7,12 +7,14 @@ import {
...
 
+import { Button } from '../buttons/Button';
 import { InputIconButton } from '../controls/InputIconButton';
 import { Box, VStack } from '../layout';
+import { Tray } from '../overlays/tray/Tray';
 
+import { Calendar, type CalendarProps, type CalendarRefHandle } from './Calendar';
 import { DateInput, type DateInputProps } from './DateInput';
 
 export type DatePickerProps = {
```
hcopp: Can we make DatePickerBaseProps (and maybe adjust web) and have all platform agnostic props be in baseprops? So styles and classNames should be in DatePickerProps since we accept different values for them depending on platform (and mobile doesn't have classNames) but most props which are shared like disabled or onClose can be in BaseProps

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -183,20 +278,38 @@ export const DatePicker = memo(
...
-              minimumDate={minDate}
-              mode="date"
-              onCancel={handleCancelNativePicker}
-              onConfirm={handleConfirmNativePicker}
-              open={showNativePicker}
-            />
+          {showPicker && (
+            <Tray
+              accessibilityRole="none"
+              footer={trayFooter}
```
hcopp: nit: this probably doesn't need to be in a useMemo? and we can use StickyFooter as well to match design exactly

```
footer={({ handleClose }) => (
            <StickyFooter background="bgElevation2" paddingX={3}>
              <Button block onPress={handleClose} ... (props from above)>
                Close
              </Button>
            </StickyFooter>
          )}
```

Grabbed from https://github.com/coinbase/cds/blob/master/packages/mobile/src/overlays/__stories__/TrayRedesign.stories.tsx#L137

Side note, I see in that example our children content has paddingBottom={1}, if we need to add bottom spacing we could do it there I suppose? Rather than paddingTop on footer. I do see we already have paddingBottom={2} for Calendar though.

packages/mobile/src/overlays/tooltip/Tooltip.tsx
```diff
@@ -26,6 +27,7 @@ export const Tooltip = memo(
     elevation,
     openDelay,
     closeDelay,
+    triggerDisabled,
```
hcopp: Can you talk through this some more? Was this just what you noticed with tooltips not being visible when the touch target it is on was disabled?

This might be a separate, more widespread change we should implement if we touch Tooltip here.

apps/docs/docs/components/other/DatePicker/mobileMetadata.json
```diff
@@ -2,29 +2,23 @@
...
     {
       "label": "Modal",
       "url": "/components/overlay/Modal/"
     }
-  ],
-  "dependencies": [
-    {
-      "name": "react-native-date-picker",
-      "version": "^4.4.2"
-    }
```
hcopp, line 28: Can you add 

```
"dependencies": [
    {
      "name": "react-native-safe-area-context",
      "version": "^4.10.5"
    }
  ]
```

Since we are using Tray? DatePicker -> Tray -> Drawer -> react-native-safe-area-context

packages/mobile/src/overlays/tooltip/Tooltip.tsx
sverg-cb: I/we actually need this anymore 😅 my 1st iteration had a different approach to the disabled calendar days, but it's much easier to use a Pressable for both enabled/disabled days and defer to how Tooltip/Pressable handle accessibility by default. It's more stable that way anyway, so I'll clean this up in the next revision.

packages/common/src/types/CalendarBaseProps.ts
```diff
@@ -0,0 +1,44 @@
...
+   * Accessibility label describing the Calendar previous month arrow.
+   * @default 'Go to previous month'
+   */
+  previousArrowAccessibilityLabel?: string;
+  /**
+   * Accessibility hint for the current day when it is not disabled. Omit or leave default for non-localized usage.
+   * @default 'Today'
+   */
+  todayAccessibilityHint?: string;
+};
```
hcopp: Sorry for any confusion on this, we don't want to include the base props in common. We will just have them both inside of Calendar.tsx and DatePicker.tsx on web and mobile but they should match.

packages/mobile/src/overlays/drawer/Drawer.tsx
```diff
@@ -105,6 +105,8 @@ export type DrawerBaseProps = SharedProps &
      * the slide transform so the drawer follows the user's finger naturally.
      */
     reduceMotion?: boolean;
+    /** Callback fired when the open animation completes. */
+    onOpenComplete?: () => void;
```
hcopp, line 109: Is this for a11y?

packages/mobile/src/overlays/tooltip/TooltipProps.ts
```diff
@@ -74,6 +74,12 @@ export type TooltipBaseProps = SharedProps &
      * is read correctly for voice-overs.
      */
     accessibilityHint?: SharedAccessibilityProps['accessibilityHint'];
+    /**
+     * Accessibility state for the trigger (e.g. disabled). Used so screen readers
+     * announce state (e.g. "dimmed" on iOS) and do not activate on double-tap;
+     * touch/press is unchanged.
+     */
+    accessibilityState?: AccessibilityState;
```
hcopp: Did we still plan on having tooltip changes? Thought those weren't needed.

packages/web/src/dates/Calendar.tsx
```diff
@@ -102,6 +103,10 @@ export type CalendarDayProps = {
   isCurrentMonth?: boolean;
   /** Tooltip content shown when hovering or focusing a disabled Calendar Day. */
   disabledError?: string;
+  /** Custom class name for the date cell (CalendarPressable). */
+  dayCellClassName?: string;
+  /** Custom style for the date cell (CalendarPressable). */
+  dayCellStyle?: React.CSSProperties;
```
hcopp: We shouldn't have style/className one offs, it should all be inside of styles/classNames

packages/web/src/dates/DatePicker.tsx
```diff
@@ -299,6 +298,7 @@ export const DatePicker = memo(
             <Calendar
               ref={calendarRef}
               className={calendarClassName}
+              classNames={classNames?.calendar}
```
hcopp: Hmm I think we have been trying to flatten styles here but please check out Stepper, Select, and Carousel as those are newer components. Also not a big deal if someone can't customize the inner workings of calendar from a root, especially if we let them override it somehow. We could look at a "CalendarComponent" prop or something.

packages/mobile/src/overlays/drawer/Drawer.tsx
sverg-cb: regarding DatePicker, yes

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,572 @@
...
+      highlighted?: StyleProp<ViewStyle>;
+      selected?: StyleProp<ViewStyle>;
+    };
+    text?: {
+      base?: StyleProp<TextStyle>;
+      selected?: StyleProp<TextStyle>;
+      disabled?: StyleProp<TextStyle>;
+      highlighted?: StyleProp<TextStyle>;
+    };
+  };
```
hcopp: We probably don't need this level of detail with dayCell and text.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,572 @@
...
+                  onPress={disableGoNextMonth ? undefined : handleGoNextMonth}
+                  style={styles?.navArrow}
+                >
+                  <Icon color="fg" name="forwardArrow" size="s" />
+                </CalendarPressable>
+              </HStack>
+            )}
+          </HStack>
+
+          <VStack style={styles?.content}>
```
hcopp: Do we need a separate VStack here? double check on web as well

sverg-cb: The VStack and HStack combination here is the closest way for us to emulate a grid layout in Mobile since CSS grids don't exist in React Native.

packages/mobile/src/overlays/tooltip/TooltipProps.ts
sverg-cb: The trigger flag is no longer needed, but adding an a11y state prop for the tooltip trigger is a way for us to guarantee the same experience for both iOS and Android w.r.t. accurately reading a "disabled" calendar day button. This is because:
* iOS VoiceOver will not read the underlying button as `dimmed` since the Tooltip's `TouchableOpacity` is not disabled, even though Android TalkBack will read the button as `disabled`.
* As a result, VoiceOver will allow SR users to toggle the tooltip content via double-tap even though the content is already read via the button's a11y hint. The content is unavailable for toggle through TalkBack.

By exposing a11y state here, we can make sure the disabled button/tooltip behaves the same for both platforms' screen readers.

packages/mobile/src/overlays/drawer/Drawer.tsx
hcopp: Was it confirmed by a11y that we want to move the focus state when tray opens? Normally it is on the close button.

Also if we do keep this we will want to add it to web Tray to keep the two platforms consistent.

sverg-cb: Spoke async—a11y expectation is to auto-focus on selected or current date when the tray opens, so this handler will be added to the Web version as well.

packages/mobile/src/dates/__stories__/DatePicker.stories.tsx
```diff
@@ -104,27 +97,50 @@ export const FullExample = () => {
...
+        />
+      </Example>
+      <Example title="DatePicker with pre-selected date (next month)">
+        <ExampleDatePicker
+          date={nextMonth15th}
+          label="Event date"
+          openCalendarAccessibilityLabel="Seed date calendar"
+        />
+      </Example>
+      <Example title="DatePicker with minimum date of today">
```
maximo-macchi-cb:
```suggestion
      <Example title="DatePicker with minimum date of tomorrow">
```

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,102 @@ export type DatePickerProps = {
...
+  /**
+   * Text to display on the confirm button.
+   * @default 'Confirm'
+   */
+  confirmText?: string;
+  /**
+   * Accessibility label for the confirm button.
+   * @default 'Confirm date selection'
+   */
+  confirmButtonAccessibilityLabel?: string;
```
maximo-macchi-cb: Since there's the `confirmText` prop, this shouldn't be needed

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,102 @@ export type DatePickerProps = {
...
+  /**
+   * Accessibility hint for the confirm button in its disabled state.
+   * Only applies when no date is selected. When a date is selected, no hint is shown.
```
maximo-macchi-cb: A11y hints are not shown on the screen
```suggestion
   * Only applies when no date is selected. When a date is selected, no hint is announced.
```

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -180,23 +271,65 @@ export const DatePicker = memo(
...
+          {showPicker && (
+            <Tray
+              accessibilityRole="none"
+              footer={({ handleClose }) => (
+                <StickyFooter paddingTop={3} paddingX={3} role="none">
+                  <Button
+                    block
+                    compact
+                    accessibilityHint={
+                      !calendarSelectedDate ? confirmButtonDisabledAccessibilityHint : undefined
```
maximo-macchi-cb: Could we just have a prop called `confirmButtonAccessibilityHint` and leave it to the customer to dynamically update the hint based on if a date is selected? I see the state of a date being selected in the calendar is kept as internal state. But it seems odd to only allow customers to add an a11y hint when the Confirm button is disabled

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -180,23 +271,65 @@ export const DatePicker = memo(
...
+                    }
+                    accessibilityLabel={confirmButtonAccessibilityLabel}
+                    disabled={!calendarSelectedDate}
```
maximo-macchi-cb: An additional prop to control the disabled state could be useful. Gives customers more granular control over how the calendar is disabled

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -180,23 +271,65 @@ export const DatePicker = memo(
...
+                    }}
+                  >
+                    {confirmText}
+                  </Button>
+                </StickyFooter>
+              )}
+              handleBarAccessibilityLabel={
+                closeCalendarAccessibilityLabel ??
+                calendarIconButtonAccessibilityLabel ??
+                'Close calendar'
```
maximo-macchi-cb: I suggest updating this so that the `closeCalendarAccessibilityLabel` prop has a default value of `'Close calendar'` and we don't fallback to `calendarIconButtonAccessibilityLabel`. `calendarIconButtonAccessibilityLabel` will be used a majority of the time to describe a button that opens the calendar. This does the opposite. The fallback of `'Close calendar'` is safer than going with the second label
```suggestion
              handleBarAccessibilityLabel={
                closeCalendarAccessibilityLabel
```

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,102 @@ export type DatePickerProps = {
...
 
 export const DatePicker = memo(
   forwardRef<View, DatePickerProps>(
     (
       {
         date,
+        styles,
+        highlightedDates,
+        nextArrowAccessibilityLabel,
+        previousArrowAccessibilityLabel,
```
maximo-macchi-cb, line 117: Can you please add defaults here?

```suggestion
        nextArrowAccessibilityLabel = 'Go to next month',
        previousArrowAccessibilityLabel = 'Go to previous month',
```

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -82,12 +148,18 @@ export const DatePicker = memo(
         invalidDateError = 'Please enter a valid date',
         disabledDateError = 'Date unavailable',
         label,
+        accessibilityHint = 'Enter date or select from calendar using the calendar button.',
         accessibilityLabel,
         accessibilityLabelledBy,
         calendarIconButtonAccessibilityLabel,
+        openCalendarAccessibilityLabel,
+        closeCalendarAccessibilityLabel,
```
maximo-macchi-cb: Please add defaults here as well

```suggestion
        openCalendarAccessibilityLabel = 'Open calendar',
        closeCalendarAccessibilityLabel = 'Close calendar without selecting a date',
```

packages/mobile/src/overlays/drawer/Drawer.tsx
maximo-macchi-cb: This behavior matches the suggested web pattern from W3 ([link](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/)). This focus behavior provides a better UX since users using ATs can quickly navigate around the calendar. The default for Drawer to focus on the close button makes sense for most Drawer instances but this is an exception.

Image

packages/mobile/src/overlays/tooltip/Tooltip.tsx
```diff
@@ -105,8 +113,12 @@ export const Tooltip = memo(
           typeof children === 'string' && accessibilityHint === undefined
             ? children
             : accessibilityHint,
+        // accessibilityState is applied to the trigger regardless of screen reader usage.
+        // Only set it when you need screen reader behavior.
+        // e.g. disabled = true: state is announced and the trigger cannot activate
```
maximo-macchi-cb, line 118: Small nit, it's already implied `accessibilityState` will always be applied so I don't think this comment is necessary

```suggestion
```

packages/mobile/package.json
```diff
@@ -187,7 +187,6 @@
     "lottie-react-native": "^6.7.0",
     "react": "^18.3.1",
     "react-native": "^0.74.5",
-    "react-native-date-picker": "^4.4.2",
```
maximo-macchi-cb, line 190: 👏 👏 👏

packages/web/src/dates/__tests__/Calendar.test.tsx
maximo-macchi-cb, line 1: There are no unit tests for web dates components (Calendar, DateInput, DatePicker)? That should be fixed. Lmk if I'm missing something here. If not, I'll create a Linear ticket to address this

packages/mobile/src/buttons/IconButton.tsx
maximo-macchi-cb, line 1: Does this reformatting happen if you try to format on `master`? Just want to confirm there isn't an issue with your Prettier setup

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,560 @@
...
+  /** Callback function fired when pressing a Calendar date. */
+  onPressDate?: (date: Date) => void;
+  /** Disables user interaction. */
+  disabled?: boolean;
+  /** Hides the Calendar next and previous month arrows, but does not prevent navigating to the next or previous months via keyboard. This probably only makes sense to be used when `minDate` and `maxDate` are set to the first and last days of the same month. */
+  hideControls?: boolean;
+  /** Array of disabled dates, and date tuples for date ranges. Make sure to set `disabledDateError` as well. A number is created for every individual date within a tuple range, so do not abuse this with massive ranges. */
+  disabledDates?: (Date | [Date, Date])[];
+  /** Array of highlighted dates, and date tuples for date ranges. A number is created for every individual date within a tuple range, so do not abuse this with massive ranges. */
+  highlightedDates?: (Date | [Date, Date])[];
```
maximo-macchi-cb: Two additions to add:
1. A prop for adding a11y hints to highlighted dates. This could be one prop applied to all highlighted dates for simplicity
2. Update this prop's type so that a customer can pass in an a11y hint for each date specifically. In case they want to add different a11y hints for different dates

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,560 @@
...
+import { useScreenReaderStatus } from '../hooks/useScreenReaderStatus';
+import { Icon } from '../icons/Icon';
+import { Box } from '../layout/Box';
+import { HStack } from '../layout/HStack';
+import { VStack, type VStackProps } from '../layout/VStack';
+import { Tooltip } from '../overlays/tooltip/Tooltip';
+import { Pressable, type PressableBaseProps } from '../system/Pressable';
+import { Text } from '../typography/Text';
+
+const CALENDAR_DAY_SIZE = 40;
```
maximo-macchi-cb: This should use a theme value

```suggestion
const CALENDAR_DAY_SIZE = 5;
```

packages/mobile/src/dates/__stories__/CalendarCustomTrigger.stories.tsx
maximo-macchi-cb, line 1: How come these examples were put in their own story?

packages/mobile/src/dates/DatePicker.tsx
sverg-cb: I can remove this! but to make sure to future reference:
If a button has text (not like the calendar days, but a simple CTA), is the a11y label usually whatever the button's text is? That should mean the screen reader announces aloud whatever is printed on the button, yeah?

packages/mobile/src/dates/Calendar.tsx
sverg-cb: In the interest of time—and since it would also need to apply to Web—we should defer to a future/KTLO issue. At least now, the mobile and web calendars are as close as they possibly can be.

packages/mobile/src/dates/__stories__/CalendarCustomTrigger.stories.tsx
sverg-cb: These can go in the "normal" Calendar story. The rest of the examples in the Calendar story all rendered calendars inline, and these examples represent something a little different since it's also about using different triggers to toggle the calendar.
I'd say the demonstration is more like DatePicker, but we shouldn't calls these date pickers since our component for that is bound to the date input as a trigger.

packages/web/src/dates/__tests__/Calendar.test.tsx
sverg-cb: That's right. In the current [master](https://github.com/coinbase/cds/tree/master/packages/web/src/dates), there is no test directory for the dates components in Web, let alone test suites for them.

packages/mobile/src/buttons/IconButton.tsx
sverg-cb: I'm intentionally adding a `forwardRef<View>` to this file! So the auto-format is intentional

packages/mobile/src/dates/DatePicker.tsx
sverg-cb:
```suggestion
                    disabled={disabled || !calendarSelectedDate}
```
Do you mean something like this? There's already a `disabled` prop added to the DatePicker, and that is for disabling the date input and calendar entirely—or are you asking for a prop specifically for when the confirm button is `disabled`?

packages/mobile/src/dates/Calendar.tsx
sverg-cb: I used 40 because the Web calendar in [master](https://github.com/coinbase/cds/blob/master/packages/web/src/dates/Calendar.tsx#L425-L427) hard-code 40 in specifically 🫠 

but for this case, are you suggesting something like this?
```tsx
const theme = useTheme();

return (
   <Box 
      height={theme.space[CALENDAR_DAY_SIZE]}
      width={theme.space[CALENDAR_DAY_SIZE]}
   />
);
```

sverg-cb: If, for example, someone changes the spacing values via `ThemeProvider` so that "40px" is actually `space[2]` or `space[5]` is 64px, do we expect the day button dimensions to also change? This would not be true in Web as every day button is a fixed 40x40px box.

packages/mobile/src/dates/DatePicker.tsx
sverg-cb: will do for the close label, but for open, I'd argue towards leaving it undefined here and defaulting further down in code.
We'll be deprecating the icon button a11y label, but it would still be used, so we would want a specific order to determine the icon button's label:
1. new prop (`openCalendarAccessibilityLabel`) if defined
2. deprecated prop (`calendarIconButtonAccessibilityLabel`) if previously defined
3. `"Open calendar"` as default

packages/mobile/src/dates/DatePicker.tsx
maximo-macchi-cb: I meant a prop specifically for disabling the Confirm Date button. Something like
```suggestion
                    disabled={confirmDateButtonDisabled || !calendarSelectedDate}
```

packages/mobile/src/dates/DatePicker.tsx
maximo-macchi-cb: Could the determination order be adjusted instead? So the order for determining the label is:
1. `calendarIconButtonAccessibilityLabel`
2. `openCalendarAccessibilityLabel`

If `openCalendarAccessibilityLabel` ends up being referenced in other places in the component later on, the default wouldn't be applied. Having the default prop values makes it less likely to avoid having empty `aria-label`'s when the new prop is used. When the deprecated prop is fully removed this determination order shouldn't break anything

packages/mobile/src/dates/Calendar.tsx
maximo-macchi-cb: Item 1 is a blocking item. Without this, there's no way for users who are blind to tell when a date they're focused on is highlighted. Depending on the workflow, this could be critical information for users. 

Even if there's no a11y hint prop in web, bugs should not be in the scope of platform parity. So while we may want to avoid adding improvements to web's DatePicker, we should address bugs on mobile that coincidentally exist in web currently.

Item 2 can definitely be a fast follow item. I understand this might take a bit of time to add and I doubt there's many cases where various highlighted dates each need their own unique a11y hint.

packages/mobile/src/dates/Calendar.tsx
maximo-macchi-cb: I'm guessing that 40px hard-coded value on web is due to a grid being used. But we should be avoiding hard-coded pixel values whenever we can. And in this case, it can be avoided since you're passing in that values to `<Box>`'s props which accept theme values. Web's behavior isn't ideal since it's ignoring the theme which our components shouldn't be doing

packages/web/src/dates/__tests__/Calendar.test.tsx
maximo-macchi-cb: Linear ticket created and in our backlog!

packages/mobile/src/dates/DatePicker.tsx
sverg-cb: The Calendar component provides default values for these already 💯

apps/docs/docs/components/other/DatePicker/_mobileExamples.mdx
```diff
@@ -187,7 +184,7 @@ function Example() {
 
 ## Styling
 
-DatePicker supports the same styling functionality as [DateInput](/components/other/DateInput/).
+DatePicker supports the same styling functionality as [DateInput](/components/other/DateInput/) for the input (`styles.dateInput`) plus [Calendar](/components/other/Calendar/) slots via `styles.calendar`.
```
hcopp: We should drop this added text or update this line.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+  width?: number;
+  height?: number;
+  background?: 'transparent' | 'bg' | 'bgPrimary';
+};
+
+const CalendarPressable = memo(
+  forwardRef<View, CalendarPressableBaseProps>(
+    ({ background = 'transparent', borderRadius = 1000, children, ...props }, ref) => {
+      const theme = useTheme();
+      const calendarDayDimensionValue = theme.space[CALENDAR_DAY_SPACING_VALUE];
```
hcopp: Ideally we don't hardcode this, maybe check v9 to see if they updated calendar when they removed hardcoded values. If it is still hardcoded for web Calendar there we can adopt the same thing here.

However we cannot hardcode to a spacing value since a dense theme here would result in the button being too small for the text. We should, ideally have it be a padding value plus the content height and that should equal 40px in our default theme.

Not sure how easy this is for text so we could just hardcode to 40px in v8 if needed. Vs an IconButton the icon is the same height and width.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+      // on some platforms a11y disabled is equivalent to the top-level disabled prop, so always
+      // setting it would block tooltip interactivity for users not using SRs.
+      const tooltipAccessibilityState = useMemo(
+        () => ({ disabled: isScreenReaderEnabled }),
+        [isScreenReaderEnabled],
+      );
+
+      if (!isCurrentMonth) {
+        const dimensionValue = theme.space[CALENDAR_DAY_SPACING_VALUE];
+        return <Box aria-hidden={true} height={dimensionValue} width={dimensionValue} />;
```
hcopp: Hmm this one is tougher, since it is blank. We could look to use a set value like 'X' and have it be 0 opacity, let me know how successful you are with the above stuff and we can discuss this.

Again, dense theme has 1/2 spacing so their buttons would be 20px by 20px with this current logic, which is not what we want.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+
+      return dayButton;
+    },
+  ),
+);
+
+CalendarDay.displayName = 'CalendarDay';
+
+/** Custom styles for individual elements of the Calendar component (mobile). */
+export type CalendarStyles = {
```
hcopp: Do we need to export this still? Ideally we don't define a separate type but rather include this in CalendarProps.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+  /** Individual date cell element, basic ViewStyle applied to the pressable wrapper */
+  day?: StyleProp<ViewStyle>;
+};
+
+export type CalendarRefHandle = {
+  /** Sets accessibility focus on the selected date, seed date, or today. */
+  focusInitialDate: () => void;
+};
+
+export type CalendarBaseProps = SharedProps & {
```
hcopp: Nit: would be nice if we included VSTackBaseProps (or BoxBaseProps) here, omitting whatever we need to omit as well. We'd want that on web and mobile.

This will be beneficial for if people customize the component using component theming (in the future) as we'd use BaseProps there.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+      const handleDatePress = useCallback(
+        (date: Date) => {
+          onPressDate?.(date);
+
+          // Note: We don't need to re-focus the button after selection.
+          // The button remains focused after activation, and both TalkBack and VoiceOver
+          // automatically announce the updated accessibilityState={{ selected: true }}.
+        },
+        [onPressDate],
+      );
```
hcopp: Do we need this callback, or could we just do `onPress={onPressDate}`

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -0,0 +1,578 @@
...
+        [disableGoNextMonth],
+      );
+
+      return (
+        <VStack
+          background="bg"
+          borderRadius={400}
+          opacity={disabled ? accessibleOpacityDisabled : undefined}
+          overflow="hidden"
+          style={styles?.root}
```
hcopp: You'll want to do `style={[style, styles?.root]}` and destructure style above.

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,101 @@ export type DatePickerProps = {
...
+   */
+  confirmText?: string;
+  /**
+   * Accessibility hint for the confirm button.
+   */
+  confirmButtonAccessibilityHint?: string;
+  /**
+   * Disables the confirm button independently of the DatePicker `disabled` prop.
+   */
+  confirmButtonDisabled?: boolean;
```
hcopp: What was the conversation behind wanting this? Without more context, I'd say someone would have just disabled the datePicker instead

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,101 @@ export type DatePickerProps = {
...
+    | 'style'
+  > &
+  Pick<
+    CalendarProps,
+    | 'seedDate'
+    | 'highlightedDates'
+    | 'highlightedDateAccessibilityHint'
+    | 'nextArrowAccessibilityLabel'
+    | 'previousArrowAccessibilityLabel'
+  >;
```
hcopp: Can some of these go on BaseProps (if CalendarBaseProps has them)

packages/web/src/overlays/tray/Tray.tsx
```diff
@@ -362,6 +366,12 @@ export const Tray = memo(
       [trayHeight, handleSwipeClose, animate, scope],
     );
 
+    const handleOpenComplete = useCallback(() => {
+      if (hasCalledOnOpenComplete.current) return;
+      hasCalledOnOpenComplete.current = true;
+      onOpenComplete?.();
+    }, [onOpenComplete]);
```
hcopp, line 373: I'm a little weary of using a ref for a callback like this, what do you think about adding

```
const [hasTriggeredOpenComplete, setHasTriggeredOpenComplete] = useState(false);

...

if (!isOpen && hasTriggeredOpenComplete) setHasTriggeredOpenComplete(false);

    const handleOpenComplete = useCallback(() => {
      if (!isOpen || hasTriggeredOpenComplete) return;
      setHasTriggeredOpenComplete(true);
      onOpenComplete?.();
    }, [hasTriggeredOpenComplete, isOpen, onOpenComplete]);

...

onAnimationComplete={handleOpenComplete}
```

packages/mobile/src/dates/DatePicker.tsx
sverg-cb: we had a small [conversation](https://github.com/coinbase/cds/pull/139#discussion_r2921179508), but on second thought...the general UX would be for someone to disable the datepicker entirely instead of _only_ the confirmation button.

i'll remove this for now. if, for whatever reason, a customer asks for this, we **could** introduce this. better than having to clean it up later on if it we learn it's not actually necessary.

packages/mobile/src/dates/Calendar.tsx
sverg-cb: The Web calendar—even in [v9](https://github.com/coinbase/cds/blob/cds-v9/packages/web/src/dates/Calendar.tsx#L425)—still uses a hardcoded 40px by 40px (x7) grid. Mobile doesn't have a grid, but fixing the cells at 40x40 would keep consistency for the near future, at the very least.

Also, since the Calendar expands and contracts relative to the screen's width (screenshots in description), I'm going to keep the day button dimensions at 40x40 for the time being.

packages/mobile/src/dates/Calendar.tsx
sverg-cb: echo from above thread: since Web still bakes in 40x40 buttons, and the Mobile calendar width corresponds to that of the screen, keeping 40x40 for the time being.

packages/web/src/overlays/tray/Tray.tsx
sverg-cb: Spoke a bit offline about this. Attaching the following summary from Claude 4.6 Opus to explain keeping the ref instead of `useState`.

Screenshot 2026-03-15 at 9 10 04 PM

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -282,11 +277,9 @@ export const DatePicker = memo(
                   <Button
                     block
                     compact
-                    accessibilityHint={
-                      !calendarSelectedDate ? confirmButtonDisabledAccessibilityHint : undefined
-                    }
-                    accessibilityLabel={confirmButtonAccessibilityLabel}
-                    disabled={!calendarSelectedDate}
+                    accessibilityHint={confirmButtonAccessibilityHint}
+                    accessibilityLabel={confirmText}
```
maximo-macchi-cb: This isn't needed since the `<Button>`'s child text content (`confirmText`) will be exposed to assistive technologies

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -24,14 +24,14 @@ import type { SharedProps } from '@coinbase/cds-common/types';
...
-import { Box } from '../layout/Box';
+import { Box, type BoxBaseProps } from '../layout/Box';
 import { HStack } from '../layout/HStack';
 import { VStack, type VStackProps } from '../layout/VStack';
 import { Tooltip } from '../overlays/tooltip/Tooltip';
 import { Pressable, type PressableBaseProps } from '../system/Pressable';
 import { Text } from '../typography/Text';
 
-const CALENDAR_DAY_SIZE = 40;
+const CALENDAR_DAY_DIMENSION = 40;
```
maximo-macchi-cb, line 34: Just confirming that we're sticking with the 40px value for now rather than a theme value. Is that correct?

packages/web/src/dates/Calendar.tsx
```diff
@@ -205,14 +214,18 @@ const CalendarDay = memo(
...
       },
       ref,
     ) => {
       const { locale } = useLocale();
       const handleClick = useCallback(() => onClick?.(date), [date, onClick]);
-      const accessibilityLabel = getDayAccessibilityLabel(date, locale);
+      const baseLabel = getDayAccessibilityLabel(date, locale);
+      const accessibilityLabel = highlighted
+        ? `${baseLabel}. ${highlightedDateAccessibilityHint}`
+        : baseLabel;
```
maximo-macchi-cb: Why is `highlightedDateAccessibilityHint` grouped into the `accessibilityLabel`? It's a small nit but I think it'd be better to pass it in separately as an `accessibilityHint` into `<CalendarPressable>`

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -540,9 +540,10 @@ export const Calendar = memo(
                       }
                       disabledError={disabledDateError}
                       highlighted={highlightedTimes.has(time)}
+                      highlightedDateAccessibilityHint={highlightedDateAccessibilityHint}
```
maximo-macchi-cb, line 526: The `highlightedDateAccessibilityHint` isn't being announced for VoiceOver when testing locally. Are you able to reproduce this?

Image

packages/web/src/dates/DatePicker.tsx
```diff
@@ -70,28 +92,45 @@ export type DatePickerProps = {
    * @default 'Open calendar' / 'Close calendar'
    */
   calendarIconButtonAccessibilityLabel?: string;
```
hcopp, line 61: Can we make this match new mobile behavior? And since this prop is on web and mobile with the same signature.

apps/docs/docs/components/other/DatePicker/_mobileExamples.mdx
```diff
@@ -149,7 +149,7 @@ function Example() {
       onErrorDate={setError}
       label="Birthdate"
       calendarIconButtonAccessibilityLabel="Open calendar to select birthdate"
-      confirmButtonAccessibilityLabel="Confirm birthdate selection"
+      confirmText="Confirm birthdate selection"
```
maximo-macchi-cb, line 155: Can you please update this example to show the usage of every a11y label prop? When I search for them I see 8 props

Image

packages/mobile/src/overlays/tooltip/TooltipProps.ts
```diff
@@ -1,4 +1,4 @@
-import type { Animated, LayoutRectangle, ViewProps } from 'react-native';
+import type { AccessibilityState, Animated, LayoutRectangle, ViewProps } from 'react-native';
```
hcopp: is this still needed? Should we have tooltipprops in here?

maximo-macchi-cb (commented):
Thank you for addressing my comments so quickly! We're getting close! :accessibility:  🚀

packages/mobile/src/dates/Calendar.tsx
sverg-cb: yeah that's the plan for now [(thread)](https://github.com/coinbase/cds/pull/139#discussion_r2931032529)

packages/web/src/dates/Calendar.tsx
sverg-cb: ah good catch 😅 trusted AI a little prematurely here, but that'll be fixed

packages/mobile/src/overlays/tooltip/TooltipProps.ts
sverg-cb: nice catch, will clean up ✨

packages/web/src/dates/Calendar.tsx
```diff
@@ -223,15 +223,13 @@ const CalendarDay = memo(
...
-      const accessibilityLabel = highlighted
-        ? `${baseLabel}. ${highlightedDateAccessibilityHint}`
-        : baseLabel;
       const calendarDayButton = useMemo(
         () => (
           <CalendarPressable
             ref={ref}
             focusable
-            accessibilityLabel={accessibilityLabel}
+            accessibilityHint={highlighted ? highlightedDateAccessibilityHint : undefined}
```
maximo-macchi-cb, line 231:
```suggestion
            accessibilityHint={highlightedDateAccessibilityHint}
```

maximo-macchi-cb (approved):
A11y approved! :accessibility: 📅 🚀

apps/docs/docs/components/other/Calendar/_mobileExamples.mdx
```diff
@@ -0,0 +1,382 @@
+Calendar is a date grid for selecting dates and powers the picker in [DatePicker](/components/other/DatePicker). It can be used standalone or inside a [Tray](/components/overlay/Tray) with a confirm action. On mobile, use the `styles` prop to customize slots; `classNames` are not supported.
```
hcopp: No need to say 'On mobile, use the `styles` prop to customize slots; `classNames` are not supported.' The rest look great.

apps/docs/docs/components/other/Calendar/_mobileExamples.mdx
```diff
@@ -0,0 +1,382 @@
...
+        navArrow: { backgroundColor: 'transparent' },
+        calendarGrid: { paddingVertical: 8 },
+        dayCell: { button: { base: { borderRadius: 8 } } },
+      }}
+    />
+  );
+}
+```
+
+On web you can also use `classNames` and CSS to target elements like icon or day-header text; on mobile only the `styles` prop is supported.
```
hcopp: Same here, no need to reference web on mobile.

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,94 @@ export type DatePickerProps = {
...
    * Accessibility label describing the calendar IconButton, which opens the calendar when pressed.
-   * @default 'Open calendar' / 'Close calendar'
+   * @deprecated Use openCalendarAccessibilityLabel/closeCalendarAccessibilityLabel instead
    */
   calendarIconButtonAccessibilityLabel?: string;
+  /**
+   * Accessibility label for the calendar IconButton, which opens the calendar when pressed.
+   * @default 'Open calendar'
+   */
+  openCalendarAccessibilityLabel?: string;
```
hcopp, line 60: Can you move `openCalendarAccessibilityLabel`, `calendarIconButtonAccessibilityLabel`, and `closeCalendarAccessibilityLabel` to BaseProps on both web and mobile?

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -7,15 +7,18 @@ import {
...
 import { InputIconButton } from '../controls/InputIconButton';
 import { Box, VStack } from '../layout';
+import { Tray } from '../overlays/tray/Tray';
+import { StickyFooter } from '../sticky-footer/StickyFooter';
 
+import { Calendar, type CalendarRefHandle } from './Calendar';
 import { DateInput, type DateInputProps } from './DateInput';
 
-export type DatePickerProps = {
+export type DatePickerBaseProps = {
```
hcopp: You can pick a decent amount of these props from Calendar, such as
```
Pick<
  CalendarBaseProps,
  | 'disabled' // this one is up to you, we could pick from DateInput
  | 'disabledDates'
  | 'minDate'
  | 'maxDate'
  | 'disabledDateError'
  | 'seedDate'
  | 'highlightedDates'
  | 'nextArrowAccessibilityLabel'
  | 'previousArrowAccessibilityLabel'
  | 'highlightedDateAccessibilityHint'
>
```

Reuse helps prevent bugs in the future as we update types and prevents inconsistencies as we update jsdocs.

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -37,39 +40,94 @@ export type DatePickerProps = {
...
+   * @default 'Confirm'
+   */
+  confirmText?: string;
+  /**
+   * Accessibility hint for the confirm button.
+   */
+  confirmButtonAccessibilityHint?: string;
+  /** Custom styles for the DateInput and Calendar subcomponents. */
+  styles?: {
+    dateInput?: DateInputProps['style'];
```
hcopp: Can we add calendar here to match web?

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -180,23 +266,59 @@ export const DatePicker = memo(
             onErrorDate={onErrorDate}
             required={required}
             requiredError={requiredError}
-            style={dateInputStyle}
+            style={dateInputStyle ?? styles?.dateInput}
```
hcopp: Hmm now that I think about it this should be the other way aroun or even better we should combine: `[dateInputStyle, styles?.dateInput]`. This way the newer prop has specificity over the older one.

packages/mobile/src/dates/DatePicker.tsx
```diff
@@ -99,75 +162,98 @@ export const DatePicker = memo(
...
-      const [showNativePicker, setShowNativePicker] = useState(false);
+      const [showPicker, setShowPicker] = useState(false);
+      const [calendarSelectedDate, setCalendarSelectedDate] = useState<Date | null>(null);
       const dateInputRef = useRef<TextInput | null>(null);
-
-      const today = useMemo(() => new Date(), []);
-
+      const calendarButtonRef = useRef<View | null>(null);
+      const calendarRef = useRef<CalendarRefHandle>(null);
+      const closedByConfirmRef = useRef(false);
```
hcopp, line 158: Can you remind me why we need a ref for this?

packages/web/src/dates/DatePicker.tsx
```diff
@@ -19,12 +19,12 @@ import {
...
 } from '../overlays/popover/PopoverProps';
 
-import { Calendar, type CalendarProps } from './Calendar';
+import { Calendar } from './Calendar';
 import { DateInput, type DateInputProps } from './DateInput';
 
 const MotionVStack = motion(VStack);
 
-export type DatePickerProps = {
+export type DatePickerBaseProps = {
```
hcopp: Same here, picking from Calendar would be a nice cleanup.

packages/web/src/dates/Calendar.tsx
sverg-cb: We should only include the highlight hint if the date is actually highlighted.

packages/mobile/src/dates/DatePicker.tsx
sverg-cb: Tray's `onCloseComplete` runs for every close (Confirm, handle bar, swipe, etc.), after the close animation—not in the same turn as the Confirm press. We must run onCancel **only** on dismiss, not after Confirm.

A `useState` flag would often still be stale by the time the callback runs since updates aren’t flushed yet. `ref.current` is updated synchronously on confirm, so when `onCloseComplete` fires we can still determine “closed via Confirm” vs “user dismissed” without changing Tray’s API.

apps/docs/docs/components/other/DatePicker/_mobileExamples.mdx
```diff
@@ -191,7 +191,7 @@ function Example() {
 
 ## Styling
 
-DatePicker supports the same styling functionality as [DateInput](/components/other/DateInput/).
+DatePicker supports the same styling functionality as [DateInput](/components/other/DateInput/) plus [Calendar](/components/other/Calendar/) via `styles.calendar`.
```
hcopp: nit: should be and Calendar (non eed to include styles.calendar since we didn't for DateInput.

hcopp: Actually this isn't really true though since we aren't exposing all the props right now, maybe we should add them, such as .calendar, .calenderHeader, ...

Otherweise wed don't support the 'same', but rather a subset.

packages/mobile/src/dates/Calendar.tsx
```diff
@@ -191,17 +191,7 @@ const CalendarDay = memo(
...
-                ? highlighted
-                  ? 'fgPrimary'
-                  : undefined
-                : active
-                  ? 'fgInverse'
-                  : highlighted
-                    ? 'fgPrimary'
-                    : undefined
-            }
+            color={active && !disabled ? 'fgInverse' : highlighted ? 'fgPrimary' : undefined}
```
hcopp, line 194: Nice

packages/ui-mobile-playground/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 4.13.1 (3/18/2026 PST)
+
+#### 🐞 Fixes
+
+- Add a route for new Calendar component. [[#139](https://github.com/coinbase/cds/pull/139)]
+
```
hcopp, line 16: No need to change this but this should be 4.14.0.

apps/docs/docs/components/other/DatePicker/_mobileExamples.mdx
sverg-cb: I'll change this to say calendar styling is root-level only—we could always exposing the remaining props should the need arise later.

PR #178, opened 2025-11-20
feat: add base props for remaining chart components

hcopp:
## What changed? Why?

This PR adds baseprops to remaining components in chart and cleans up a few misc type inconsistencies.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/mobile-visualization/src/chart/bar/Bar.tsx
+++ b/packages/mobile-visualization/src/chart/bar/Bar.tsx
@@ -5,7 +5,7 @@ import { getBarPath, type Transition } from '../utils';
 
 import { DefaultBar } from './DefaultBar';
 
-export type BarComponentProps = {
+export type BarBaseProps = {
   /**
    * X coordinate of the bar (left edge).
    */
@@ -23,7 +23,8 @@ export type BarComponentProps = {
    */
   height: number;
   /**
-   * Border radius of the bar.
+   * Border radius for the bar.
+   * @default 4
    */
   borderRadius?: number;
   /**
@@ -47,10 +48,6 @@ export type BarComponentProps = {
    */
   dataY?: number | [number, number] | null;
   /**
-   * The path data for the bar shape.
-   */
-  d: string;
-  /**
    * Fill color for the bar.
    */
   fill?: string;
@@ -67,26 +64,26 @@ export type BarComponentProps = {
    */
   strokeWidth?: number;
   /**
-   * Transition configuration for bar animations.
-   */
-  transition?: Transition;
-};
-
-export type BarComponent = React.FC<BarComponentProps>;
-
-export type BarBaseProps = Omit<BarComponentProps, 'd' | 'transition'> & {
-  /**
-   * Border radius for the bar.
-   * @default 4
-   */
-  borderRadius?: BarComponentProps['borderRadius'];
-  /**
    * Component to render the bar.
    */
   BarComponent?: BarComponent;
 };
 
-export type BarProps = BarBaseProps & Pick<BarComponentProps, 'transition'>;
+export type BarProps = BarBaseProps & {
+  /**
+   * Transition configuration for bar animations.
+   */
+  transition?: Transition;
+};
+
+export type BarComponentProps = Omit<BarProps, 'BarComponent'> & {
+  /**
+   * The path data for the bar shape.
+   */
+  d: string;
+};
+
+export type BarComponent = React.FC<BarComponentProps>;
 
 /**
  * Simple bar component that renders a single bar at the specified position.
--- a/packages/mobile-visualization/src/chart/line/Line.tsx
+++ b/packages/mobile-visualization/src/chart/line/Line.tsx
@@ -21,56 +21,7 @@ import { convertToSerializableScale } from '../utils/scale';
 import { DottedLine } from './DottedLine';
 import { SolidLine } from './SolidLine';
 
-export type LineComponentProps = {
-  /**
-   * Path of the line
-   */
-  d: AnimatedProp<string | undefined>;
-  /**
-   * The color of the line.
-   * @default color of the series or theme.color.fgPrimary
-   */
-  stroke: string;
-  /**
-   * Opacity of the line
-   * @note when combined with gradient, both will be applied
-   * @default 1
-   */
-  strokeOpacity?: number;
-  /**
-   * Width of the line
-   * @default 2
-   */
-  strokeWidth?: number;
-  /**
-   * ID of the y-axis to use.
-   * If not provided, defaults to the default y-axis.
-   */
-  yAxisId?: string;
-  /**
-   * Gradient configuration.
-   * When provided, creates gradient or threshold-based coloring.
-   */
-  gradient?: GradientDefinition;
-  /**
-   * Whether to animate the line.
-   * Overrides the animate value from the chart context.
-   */
-  animate?: boolean;
-  /**
-   * Transition configuration for line animations.
-   */
-  transition?: Transition;
-};
-
-export type LineComponent = React.FC<LineComponentProps>;
-
-export type LineBaseProps = Partial<
-  Pick<
-    LineComponentProps,
-    'stroke' | 'strokeWidth' | 'strokeOpacity' | 'gradient' | 'animate' | 'transition'
-  >
-> & {
+export type LineBaseProps = {
   /**
    * The ID of the series to render. Will be used to find the data from the chart context.
    */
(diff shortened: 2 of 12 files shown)


packages/mobile-visualization/src/chart/line/Line.tsx
```diff
@@ -66,73 +65,74 @@ export type LineComponentProps = {
 
 export type LineComponent = React.FC<LineComponentProps>;
 
-export type LineProps = Partial<
+export type LineBaseProps = Partial<
   Pick<
     LineComponentProps,
     'stroke' | 'strokeWidth' | 'strokeOpacity' | 'gradient' | 'animate' | 'transition'
```
hcopp: We should have transition be on LineProps, not baseProps

packages/mobile-visualization/src/chart/line/Line.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { memo, useEffect, useMemo } from 'react';
 import { useSharedValue, withDelay, withTiming } from 'react-native-reanimated';
-import type { SharedProps } from '@coinbase/cds-common/types';
```
hcopp, line 3: I am dropping SharedProps for every component inside of Skia's Canvas since there is no way to set testIDs on those components & retrieve it in Jest

packages/mobile-visualization/src/chart/scrubber/Scrubber.tsx
```diff
@@ -133,8 +133,10 @@ export type ScrubberBeaconLabelComponent = React.FC<ScrubberBeaconLabelProps>;
 export type ScrubberLabelProps = ReferenceLineLabelComponentProps;
 export type ScrubberLabelComponent = React.FC<ScrubberLabelProps>;
 
-export type ScrubberBaseProps = SharedProps &
-  Pick<ScrubberBeaconGroupBaseProps, 'idlePulse'> & {
+export type ScrubberBaseProps = Pick<ScrubberBeaconGroupBaseProps, 'idlePulse'> &
+  Pick<ReferenceLineBaseProps, 'LineComponent' | 'LabelComponent' | 'elevateLabel'> &
+  Pick<ScrubberBeaconGroupProps, 'BeaconComponent'> &
+  Pick<ScrubberBeaconLabelGroupProps, 'BeaconLabelComponent'> & {
```
hcopp, line 139: I am keeping the different component props (such as `BeaconComponent` and `BeaconLabelComponent`) on BaseProps since that is what we have done on other components

packages/mobile-visualization/src/chart/Path.tsx
```diff
@@ -20,7 +20,61 @@ import { unwrapAnimatedValue } from './utils';
...
+   *
+   * @example
+   * // Duration based
+   * transition={{ type: 'timing', duration: 300 }}
+   *
+   * @example
+   * // Spring based
+   * transition={{ type: 'spring', damping: 20, stiffness: 300 }}
+   */
+  transition?: Transition;
```
hcopp: Move to PathProps since the object is different on web and mobile.

packages/mobile-visualization/src/chart/Path.tsx
```diff
@@ -37,67 +91,16 @@ export type PathProps = SharedProps &
...
-     * Custom clip path rect. If provided, this overrides the default chart rect for clipping.
-     * Will be overridden by clipPath if set.
-     */
-    clipRect?: Rect;
     /**
      * Custom clip path.
      * When set, overrides clipRect.
      * @note pass undefined to disable clipping.
      */
     clipPath?: string | undefined;
```
hcopp: Don't need undefined here, maybe we need to accept null instead of undefined - and maybe in some other places as well we can support no component by doing this.

packages/web-visualization/src/chart/line/ReferenceLine.tsx
hcopp, line 119: Can we add these instead at ReferenceLineProps

yarn.lock
```diff
@@ -2428,7 +2428,7 @@ __metadata:
     "@coinbase/cds-lottie-files": "workspace:^"
     "@coinbase/cds-mobile": "workspace:^"
     "@coinbase/cds-utils": "workspace:^"
-    "@shopify/react-native-skia": 1.12.4
+    "@shopify/react-native-skia": ^1.12.4 || ^2.0.0
```
hcopp, line 2431: Slipping this in - didn't run yarn recently. We support v1 or v2 since retail-native already has v2. However we can't use v2 ourselves since you need to be on React 19.

packages/mobile-visualization/src/chart/area/AreaChart.tsx
```diff
@@ -82,6 +86,9 @@ export type AreaChartProps = Omit<CartesianChartProps, 'xAxis' | 'yAxis' | 'seri
     yAxis?: Partial<AxisConfigProps> & YAxisProps;
   };
 
+export type AreaChartProps = AreaChartBaseProps &
+  Omit<CartesianChartProps, 'xAxis' | 'yAxis' | 'series'>;
```
cb-ekuersch, line 90: Above you omitted theese same fields from the Base props

packages/mobile-visualization/src/chart/axis/YAxis.tsx
```diff
@@ -211,7 +211,7 @@ export const YAxis = memo<YAxisProps>(
                 <TickMarkLineComponent
                   key={`tick-mark-${tick.tick}-${index}`}
                   animate={false}
-                  clipPath={undefined}
+                  clipPath={null}
```
cb-ekuersch, line 214: does the component handle undefined & null diffrently?

packages/mobile-visualization/src/chart/bar/Bar.tsx
```diff
@@ -66,11 +66,15 @@ export type BarComponentProps = {
...
+  /**
+   * Transition configuration for bar animations.
+   */
+  transition?: Transition;
 };
 
 export type BarComponent = React.FC<BarComponentProps>;
 
-export type BarProps = Omit<BarComponentProps, 'd'> & {
+export type BarBaseProps = Omit<BarComponentProps, 'd' | 'transition'> & {
```
cb-ekuersch: i wonder if we should consider omittiing/picking from normal props in a base props definition a code smell? To that point, could you use the base props here instead?

packages/mobile-visualization/src/chart/line/Line.tsx
cb-ekuersch: nit: could be useful to have AI do a little writeup about skia and its limitations in a local README or something

packages/mobile-visualization/src/chart/Path.tsx
```diff
@@ -37,55 +79,16 @@ export type PathProps = SharedProps &
...
      * When set, overrides clipRect.
-     * @note pass undefined to disable clipping.
+     * @note pass null to disable clipping.
      */
-    clipPath?: string | undefined;
-    /**
-     * The offset to add to the clip rect boundaries.
-     */
-    clipOffset?: number;
+    clipPath?: string | null;
```
cb-ekuersch, line 86: do you want to remove the optionality here to make it more explicity on how to use?

yarn.lock
cb-ekuersch: hmm shouldn't this be in the package.json now the yarn lock?

cb-ekuersch: peer dependency info?

hcopp: It is a peerDependency in our charts feature branch https://github.com/coinbase/cds/blob/hunter/charts-beta-4/packages/mobile-visualization/package.json#L39 and the devDependency is exactly 1.12.4

packages/mobile-visualization/src/chart/axis/YAxis.tsx
hcopp: So for web we don't have clipPath it is clipRect and it handles it by if a value is undefined then it will clip to the drawingArea but if it is defined it will use that value, but null equals no clipping at all.

On mobile we were looking to see if the value of clipPath was set - and if it was set we would use that value, so setting clipPath={undefined} is what you needed to prevent the default behavior of clipping to drawing area.

I think using null is a cleaner solution for preventing clip, let me know what you think.

packages/mobile-visualization/src/chart/Path.tsx
hcopp: If you don't set clipPath it does have default clip to drawingArea, let me clarify in the JSdocs

packages/mobile-visualization/src/chart/bar/Bar.tsx
hcopp: Good callout.

Our sub components (such as DefaultSlideButtonHandle or DefaultBar with SlideButtonHandleProps or BarComponentProps) never seem to have baseprops

https://github.com/coinbase/cds/blob/master/packages/mobile/src/buttons/SlideButton.tsx#L26
https://github.com/coinbase/cds/blob/master/packages/web/src/alpha/select/Select.tsx#L119

I will look at swapping this so that BarBaseProps have everything and then BarComponentProps Picks from BarProps

packages/mobile-visualization/src/chart/area/AreaChart.tsx
hcopp: I think we would need to do it for both right? otherwise we will be getting th eprops in AreaChartProps since CartesianChartProps extends CartesianChartBaseProps

PR #182, opened 2025-11-20
feat: `<Combobox>` and `<Tray>` header + footer

maximo-macchi-cb:
## What changed? Why?
This PR introduces a new web and mobile component: `<Combobox>`. `<Combobox>` is based off of `<Select>` and adds search functionality. Some high level additions include:

- Custom search functionality via prop API
- Single / multi select support
- Added `contentNode` to `SelectControlComponent` props. This is used to render the search input below the control's selected value
- Bug fix where checkboxes / radios in mobile dropdown weren't selected / deselecting items

This PR also adds `header` props to web and mobile `<Tray>` and adds a `footer` prop to the mobile `<Tray>`. This creates more parity between the two components and is used to build `<Combobox>` on mobile.

## UI changes

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/common/CHANGELOG.md
+++ b/packages/common/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.29.0 ((12/12/2025, 01:12 PM PST))
+
+This is an artificial version bump with no new change.
+
 ## 8.28.2 ((12/12/2025, 06:23 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-common",
-  "version": "8.28.2",
+  "version": "8.29.0",
   "description": "Coinbase Design System - Common",
   "repository": {
     "type": "git",
--- a/packages/mcp-server/CHANGELOG.md
+++ b/packages/mcp-server/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.29.0 ((12/12/2025, 01:12 PM PST))
+
+This is an artificial version bump with no new change.
+
 ## 8.28.2 ((12/12/2025, 06:23 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/mcp-server/package.json
+++ b/packages/mcp-server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-mcp-server",
-  "version": "8.28.2",
+  "version": "8.29.0",
   "description": "Coinbase Design System - MCP Server",
   "repository": {
     "type": "git",
--- a/packages/mobile/CHANGELOG.md
+++ b/packages/mobile/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.29.0 (12/12/2025 PST)
+
+#### 🚀 Updates
+
+- Combobox component and header + footer props for Tray. [[#182](https://github.com/coinbase/cds/pull/182)]
+
 ## 8.28.2 (12/12/2025 PST)
 
 #### 🐞 Fixes
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-mobile",
-  "version": "8.28.2",
+  "version": "8.29.0",
   "description": "Coinbase Design System - Mobile",
   "repository": {
     "type": "git",
--- a/packages/web/CHANGELOG.md
+++ b/packages/web/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 
 <!-- template-start -->
 
+## 8.29.0 (12/12/2025 PST)
+
+#### 🚀 Updates
+
+- Combobox component and header prop for Tray. [[#182](https://github.com/coinbase/cds/pull/182)]
+
 ## 8.28.2 ((12/12/2025, 06:23 AM PST))
 
 This is an artificial version bump with no new change.
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/cds-web",
-  "version": "8.28.2",
+  "version": "8.29.0",
   "description": "Coinbase Design System - Web",
   "repository": {
     "type": "git",


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/overlays/tray/Tray.tsx
```diff
@@ -28,7 +28,9 @@ export type TrayRenderChildren = React.FC<{ handleClose: () => void }>;
 
 export type TrayBaseProps = {
   children: React.ReactNode | TrayRenderChildren;
-  /** Optional footer content that will be fixed to the bottom of the tray */
+  /** ReactNode to render as the Drawer header */
+  header?: React.ReactNode;
+  /** ReactNode to render as the Drawer footer */
```
hcopp, line 33:
```suggestion
  /** ReactNode to render as the Tray footer */
```

Should this say Tray instead of drawer?

packages/mobile/src/overlays/tray/Tray.tsx
```diff
@@ -22,6 +22,10 @@ export type TrayRenderChildren = React.FC<{ handleClose: () => void }>;
 
 export type TrayBaseProps = Omit<DrawerBaseProps, 'pin' | 'children'> & {
   children: React.ReactNode | TrayRenderChildren;
+  /** ReactNode to render as the Drawer header */
```
cb-ekuersch: nit: maybe worth extending the documentation/storybook of Tray to include an example using these.

packages/web/src/overlays/tray/Tray.tsx
```diff
@@ -28,7 +28,9 @@ export type TrayRenderChildren = React.FC<{ handleClose: () => void }>;
 
 export type TrayBaseProps = {
   children: React.ReactNode | TrayRenderChildren;
-  /** Optional footer content that will be fixed to the bottom of the tray */
+  /** ReactNode to render as the Drawer header */
+  header?: React.ReactNode;
```
cb-ekuersch, line 32: thanks for keeping things consistent with mobile! Do you think we should use Tray on web for mobile screen sizes?

packages/mobile/src/alpha/select/Select.tsx
```diff
@@ -193,6 +196,7 @@ export type SelectDropdownProps<
 > = SelectState<Type, SelectOptionValue> &
   Pick<SharedAccessibilityProps, 'accessibilityLabel'> &
   Omit<BoxProps, 'onChange'> &
+  Pick<TrayProps, 'header' | 'footer'> &
```
cb-ekuersch: nit: picking header and footer is the right move generally, but I feel like the names don't really make sense from the perspective of the Select component. e.g. `optionTrayHeader: TrayProps["header"]`?

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+export type ComboboxControlComponent<
+  Type extends SelectType = 'single',
+  SelectOptionValue extends string = string,
+> = React.FC<ComboboxControlProps<Type, SelectOptionValue> & { ref?: React.Ref<HTMLElement> }>;
+
+export type ComboboxBaseProps<
+  Type extends SelectType = 'single',
+  SelectOptionValue extends string = string,
+> = SelectBaseProps<Type, SelectOptionValue> & {
```
cb-ekuersch, line 92: love this composition!

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,240 @@
...
+      return (
+        <Select
+          ref={controlRef}
+          SelectControlComponent={ComboboxControlComponent}
+          SelectDropdownComponent={ComboboxDropdownComponent}
+          accessibilityLabel={accessibilityLabel}
+          defaultOpen={defaultOpen}
+          disabled={disabled}
+          label={label}
+          onChange={(value) => onChange?.(value)}
```
cb-ekuersch: nit: hmm should we memoize this?

packages/mobile/src/alpha/select/DefaultSelectControl.tsx
```diff
@@ -206,6 +208,7 @@ export const DefaultSelectControlComponent = memo(
             style={[{ flexGrow: 1 }, styles?.controlInputNode]}
             {...props}
           >
+            {/* TO DO: Why is this so different than the web implementation? */}
```
hcopp: @maximo-macchi-cb you have a todo here

packages/web/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,203 @@
...
+                  }}
+                  tabIndex={0}
+                  value={searchTextRef.current}
+                />
+              }
+              options={optionsRef.current}
+              placeholder={null}
+              styles={{
+                controlEndNode: { alignItems: hasValue ? 'flex-end' : 'center' },
+              }}
```
hcopp: I seem to be running into a keyboard navigation issue. When I initially tab into the component it focuses on the outer component (maybe the button from the control component). I then need to tab again to activate this input for me to type.

In addition, if I have a value that is not alphabetical (such as !) it won't open the dropdown since we are searching for A-Z

https://github.com/user-attachments/assets/a13d8052-281a-4d09-b045-f0a2f40f1bca

packages/web/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,203 @@
...
+        );
+
+      const [openInternal, setOpenInternal] = useState(defaultOpen ?? false);
+      const open = openProp ?? openInternal;
+      const setOpen = setOpenProp ?? setOpenInternal;
+
+      if (
+        (typeof openProp === 'undefined' && typeof setOpenProp !== 'undefined') ||
+        (typeof openProp !== 'undefined' && typeof setOpenProp === 'undefined')
+      )
```
hcopp: minor: could simplify by doing `(typeof openProp === 'undefined') !== (typeof setOpenProp === 'undefined')`

packages/mobile/src/alpha/select/Select.tsx
```diff
@@ -193,6 +196,7 @@ export type SelectDropdownProps<
...
   Pick<SelectOptionProps<Type, SelectOptionValue>, 'accessory' | 'media' | 'end'> & {
     /** Whether this is for single or multi-select */
     type?: Type;
```
hcopp: This might have been from your OG select pr but I'm seeing a circular import warning on mobile

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,247 @@
...
+                <StickyFooter background="bg">
+                  <Button compact onPress={() => setOpen(false)}>
+                    {closeButtonLabel}
+                  </Button>
+                </StickyFooter>
+              </KeyboardAvoidingView>
+            }
+            header={
+              <Box paddingX={3}>
+                <ComboboxControlComponent {...props} label={null} styles={undefined} />
```
hcopp: When I click the edge of this component inside of the tray it closes, rather than letting me type. I suspect this is due to me selecting the control component which toggles open state rather than selecting the NativeInput. Sounds similar to the web issue I'm seeing below.

https://github.com/user-attachments/assets/95b34717-13d8-4347-abb7-5227b29a9148

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,247 @@
...
+      // when the filtered options change. Recreating the dropdown would
+      // cause the tray inside to close and reopen
+      const filteredOptionsRef = useRef(filteredOptions);
+      filteredOptionsRef.current = filteredOptions;
+
+      const ComboboxDropdownComponent = useCallback(
+        (props: SelectDropdownProps<Type, SelectOptionValue>) => (
+          <SelectDropdownComponent
+            label={label}
+            minHeight={500}
```
hcopp, line 274: @maximo-macchi-cb is there a need for this? I'm seeing the use of KeyboardAvoidingView below, but removing this minHeight the keyboard avoiding view doesn't seem to increase in height to help out.

I'm curious what happens on a small device, if you are able to try it out with one of the old iPhone mini simulators

packages/web/src/overlays/tray/Tray.tsx
maximo-macchi-cb: Do you mean when you're on a web browser on a mobile device? That would use the web component regardless. But the styles should still work for those smaller screen sizes as long as the slotted content is styled appropriately

packages/mobile/src/alpha/select/Select.tsx
maximo-macchi-cb: Fair point! These props aren't being surfaced to `SelectBaseProps`. So you can't use it like this
```
<Select header={<Text>Hello world</Text>} />
```

It can only be used by the `SelectDropdownComponent` you pass in. So it can only be used like this, the way `<Combobox>` does
```
<SelectDropdownComponent
  header={<SelectControlComponent>
  footer={<StickyFooterDoneButton>}
>
```

With that usage above, I think it makes sense to a customer since the props are directly applied to the dropdown. It's implied that it's the header / footer of the dropdown.

packages/mobile/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: Yeah, as it seems you saw testing locally without the min height the `<KeyboardAvoidingView>` doesn't increase the height of the tray. This felt like a safe workaround to cover the edge case of having just a few options. It only creates extra white space if you have less than ~9 options (depending on your screen's height). If you have such few options, a regular `<Select>` would probably work better since there isn't a need to search. But this helps ensure that if someone insists on using `<Combobox>` with very few options, the `<KeyboardAvodingView>` still shifts the Done button up correctly.

I've attached screenshots of how it looks on an iPhone 16 Pro without the `minHeight` set. And how it looks on an iPhone SE with the `minHeight` set. 

Screenshot 2025-11-25 at 5 08 12 PM
Screenshot 2025-11-25 at 5 12 06 PM
Screenshot 2025-11-25 at 5 12 00 PM

packages/mobile/src/overlays/tray/Tray.tsx
maximo-macchi-cb: I think it'd be better to save this for its own PR just to keep the review scope smaller!

packages/mobile/src/alpha/select/Select.tsx
maximo-macchi-cb: This was due to `<DefaultSelectDropdown>` being passed into the `<Select>`'s `SelectControlComponent` prop used in `<Combobox>`. The `defaultAccessibilityRoles` are defined in `Select.tsx` and imported into `DefaultSelectDropdown.tsx`. I created a `defaultAccessibilityRoles` constant for the `DefaultSelectDropdown.tsx` files to resolve this

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+      // Store in refs to avoid recreating ComboboxControl on every search text change
+      const searchTextRef = useRef(searchText);
+      searchTextRef.current = searchText;
+      const setSearchTextRef = useRef(setSearchText);
+      setSearchTextRef.current = setSearchText;
+      const valueRef = useRef(value);
+      valueRef.current = value;
+      const optionsRef = useRef(options);
+      optionsRef.current = options;
```
hcopp: @maximo-macchi-cb did you discuss using refs with Cody? Trying to wrap my head around this and why it is needed. It might be signaling an issue with how we define ComboboxControlComponent

packages/web/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,234 @@
...
+      ref: React.Ref<ComboboxRef>,
+    ) => {
+      const [searchTextInternal, setSearchTextInternal] = useState(defaultSearchText);
+      const searchText = searchTextProp ?? searchTextInternal;
+      const setSearchText = onSearchProp ?? setSearchTextInternal;
+
+      if (
+        (typeof searchTextProp === 'undefined' && typeof onSearchProp !== 'undefined') ||
+        (typeof searchTextProp !== 'undefined' && typeof onSearchProp === 'undefined')
+      )
```
hcopp: nit: you'd also be able to simplify searchTextProp check like we did for openProp

packages/mobile/src/alpha/combobox/__stories__/Combobox.stories.tsx
```diff
@@ -0,0 +1,767 @@
...
+    <Combobox
+      hideSearchInput
+      label="Hide search input"
+      onChange={onChange}
+      options={fruitOptions}
+      placeholder="Search..."
+      type="multi"
+      value={value}
+    />
+  );
```
hcopp, line 200: Image

Is this expected to have no bottom padding?

hcopp: Overall I'm still not experiencing what I would expect as a user when interacting with the tray. When I click on the top portion of the input, where the chips are, it closes the tray. When I would have expected it to let me type. It feels weird to me to click anywhere on the control element that is inside of the tray and have it close the tray instead of toggling the focus of the text input.

maximo-macchi-cb: I can check with design but I believe the only touch target that should bring focus to the text input is the text input itself. Meaning if you touch anywhere outside the text input, it won't focus and bring up the keyboard for typing

packages/mobile/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: I updated the comment above the refs to clarify their need. The types for `SelectControlComponent` don't allow additional props to be added to the component passed in. However, the `ComboboxControlComponent` needs search props (`searchText`, `onSearch`, etc.) to function correctly. This usage of refs prevents the control from being re-rendered. If it's re-rendered, it causes rendering bugs with the dropdown menus for both web and mobile. The dropdown menus need a stable control component ref to attach to. 

@cb-ekuersch and I paired on this and I've attempted a few solutions to get the types working so that an inline component definition with refs could be avoided. However, we couldn't come to a better solution than this one.

packages/mobile/src/alpha/combobox/__stories__/Combobox.stories.tsx
maximo-macchi-cb: The behavior of this has been updated so that the text input is hidden when values are selected. For now, the touch target behavior will remain as is

packages/mobile/src/alpha/select/__stories__/AlphaSelect.stories.tsx
```diff
@@ -194,21 +194,6 @@ const CompactExample = () => {
...
-      label="Single select - label variant inside"
-      labelVariant="inside"
-      onChange={setValue}
-      options={exampleOptions}
-      placeholder="Empty value"
-      value={value}
-    />
-  );
-};
-
```
hcopp, line 211: Any reason for getting rid of this example?

packages/mobile/src/alpha/select/DefaultSelectControl.tsx
```diff
@@ -153,21 +152,14 @@ export const DefaultSelectControlComponent = memo(
...
             >
-              <InputLabel
-                alignSelf={labelVariant === 'inside' ? 'flex-start' : undefined}
-                color="fg"
-                ellipsizeMode="tail"
-                numberOfLines={2}
-                paddingX={labelVariant === 'inside' ? 2 : 0}
-                paddingY={shouldShowCompactLabel || labelVariant === 'inside' ? 0 : 0.5}
-              >
+              <InputLabel color="fg" paddingX={0} paddingY={0.5}>
```
hcopp: Are we dropping support for labelVariant?

hcopp: Image

(this is web)

packages/web/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,243 @@
...
+      const searchTextRef = useRef(searchText);
+      searchTextRef.current = searchText;
+      const setSearchTextRef = useRef(setSearchText);
+      setSearchTextRef.current = setSearchText;
+      const valueRef = useRef(value);
+      valueRef.current = value;
+      const optionsRef = useRef(options);
+      optionsRef.current = options;
+      const openRef = useRef(open);
+      openRef.current = open;
```
hcopp: @maximo-macchi-cb can you give a rundown on what you found during your searching on what would be possible vs not here?

packages/mobile/src/alpha/select/__stories__/AlphaSelect.stories.tsx
maximo-macchi-cb: Yes we're removing support for the label variant. I had to sync with the designers on this but there isn't support for inside labels for Select or Combobox

packages/mobile/src/alpha/select/DefaultSelectControl.tsx
maximo-macchi-cb: Yes we're removing support for the label variant. I had to sync with the designers on this but there isn't support for inside labels for Select or Combobox. I'll remove those web examples

packages/web/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: `<Combobox>` uses a `<Select>` under the hood and one of Select's props is `SelectControlComponent`. This prop is being used to render `<Combobox>`. Combobox is like a Select with a different control and some additional props to allow customers to modify Combobox specific behavior like search. 

`SelectControlComponent`'s prop type is `SelectControlComponent`. This prop type doesn't support the Combobox specific props needed (e.g. `searchText`). When attempting to define a default component to extend the `SelectControlComponent` prop type like this
```
type ComboboxControlProps  = SelectControlProps & {
  /** Search text value */
  searchText: string;
  /** Search text change handler */
  onSearch: (searchText: string) => void;
};
```
there are type errors thrown due to the type restrictions for the `SelectControlComponent` prop. The prop types for `ComboboxControlProps` and `SelectControlProps` don't match so the `SelectControlComponent` prop can't accept this as a parameter. 

By using the inline memoized component, the control component passed in doesn't have to add additional props so the type of `SelectControlProps` can still be satisfied. The information that would be read from the added props comes from the refs added instead.

`<Select>` could potentially be refactored to change how `SelectControlComponent` prop works so that it accepts a render method rather than a component type. However, this could be a larger refactor and cause potential issues for customers using the new `<Select>` today.

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,361 @@
...
+  options: SelectOptionList<Type, SelectOptionValue>;
+  searchText: string;
+  onSearch: (searchText: string) => void;
+  hideSearchInput: boolean;
+};
+
+/**
+ * Context used for Combobox props needed to render to the ComboboxControlComponent.
+ */
+const ComboboxContext = createContext<ComboboxContextValue<any, any> | null>(null);
```
cb-ekuersch, line 50: @maximo-macchi-cb we should probably leave a note any time we are forced to use `any` which im guessing is the case here. What happened when you tried to do: ComboboxContextValue<SelectType, string>

cb-ekuersch: and maybe `unknown` is the proper type to use in this case

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,361 @@
...
+    /** Reference to the search input */
+    searchInputRef: React.RefObject<TextInput>;
+    /** Reference to the combobox control for positioning */
+    controlRef: React.RefObject<ComboboxRef | null>;
+    /** Custom SelectControlComponent to wrap */
+    SelectControlComponent?: SelectControlComponent<Type, SelectOptionValue>;
+  };
+
+export type ComboboxControlComponent = <
+  Type extends SelectType = 'single',
```
cb-ekuersch, line 83: is there a reason this explicitly defaults to single instead of keeping the string union type?

packages/mobile/src/alpha/combobox/Combobox.tsx
```diff
@@ -0,0 +1,361 @@
...
+  SelectOptionValue extends string = string,
+>(
+  props: ComboboxControlProps<Type, SelectOptionValue>,
+) => React.ReactElement;
+
+type ComboboxControlWrapperType = <
```
cb-ekuersch: nit: lets move this so its right above the actual Wrapper.

Let's also give this pattern a more accurate name than wrapper. Maybe ComboboxControlSelectAdapter?

packages/mobile/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: This was part of the work done with `<Select>`. If the `type` prop is undefined (and this prop is optional so it's possible), the value of `Type` cannot be inferred. `<Combobox>` can be either single or multi select so we need to account for this

packages/mobile/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: I went with `ComboboxControlContextAdapter` since it's purpose is to adapt the Combobox context and pass it into the ComboboxControlComponent's props

packages/mobile/src/alpha/combobox/Combobox.tsx
maximo-macchi-cb: Updated the comment on why `any` must be used and why `unknown` won't work.

cb-ekuersch (dismissed):

cb-ekuersch (approved):

PR #253, opened 2025-12-16
feat: Add static class names to web components

stacysun-cb:
## What changed? Why?
Add static class names to the top level of the following web components to allow customization. 

If a component has a conditional wrapper, we currently apply static class names only to the wrapper (for example, `Control` and `Switch`). When no conditional wrapper is applied on `Switch`/`Control`, you can still target the elements using the `interactable` class name (see screenshot). This behavior will be improved in v9, where we will always render a wrapper component and consistently apply the static class name.

- Cell
- Collapsible
- Control
- Switch
- IconButton
- Icon
- Spinner
- Avatar
- HexagonBorder
- RemoteImage
- NavigationBar
- Interactable
- ListCell

Notes: For components that render another component, this may result in multiple static class names on a single DOM element. This is a deliberate trade-off to make components easier and more predictable to target.

## Static classnames examples
Screenshot 2025-12-16 at 3 36 04 PM
Screenshot 2025-12-16 at 3 36 37 PM
Screenshot 2025-12-16 at 3 37 04 PM

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/media/Hexagon.tsx
```diff
@@ -101,7 +101,7 @@ export const HexagonBorder = memo(
     return (
       <Box
         aria-hidden
-        className={cx(hexagonBorderContainerCss, className)}
+        className={cx('hexagon-border', hexagonBorderContainerCss, className)}
```
cb-ekuersch: this feels more like a sub-element of avatar doesn't it? At least that's the only place where i know this is used

packages/web/src/controls/Control.tsx
```diff
@@ -189,7 +189,7 @@ const ControlWithRef = forwardRef(function ControlWithRef<T extends string>(
      */
     if (!label) return iconElement;
     return (
-      <label className={pointerCss} htmlFor={inputId} style={labelStyle}>
+      <label className={cx('control', pointerCss)} htmlFor={inputId} style={labelStyle}>
```
cb-ekuersch: this looks like it gets applied conditionally

packages/web/src/controls/Switch.tsx
```diff
@@ -112,6 +112,7 @@ const SwitchWithRef = forwardRef<HTMLInputElement, SwitchProps>(function SwitchW
   return children ? (
     <Box
       alignItems="center"
+      className="switch"
```
cb-ekuersch: appears conditionally

packages/web/src/controls/Control.tsx
stacysun-cb: yes, but the inner iconElement has `interactable` classname

packages/web/src/controls/Switch.tsx
stacysun-cb: Same here - the inner `switchNode` children has `control` classname

stacysun-cb: Updated to always apply the class name at the top level. On reconsideration, this makes it easier to target.

packages/web/src/media/Hexagon.tsx
stacysun-cb: removed on sub components

packages/web/src/buttons/IconButton.tsx
```diff
@@ -12,6 +12,8 @@ import { Pressable, type PressableBaseProps } from '../system/Pressable';
 
 import { type ButtonBaseProps, spinnerHeight } from './Button';
 
+const COMPONENT_STATIC_CLASSNAME = 'icon-button';
+
```
hcopp, line 16: Thanks for doing this!

I've seen other libraries have a prefix for their static component classes, such as [.mantine-Button-root](https://mantine.dev/core/button/?t=styles-api) and [.MuiButton-root](https://mui.com/material-ui/api/button/). 

We don't have styles and classNames props shared across the board but it would be cool if we took a similar approach here and aligned with a prefix for CDS and shared the specific className key we would normally use for classNames.

Maybe something like `cdsIconButton-root`, `cds-iconButton-root`, or `cds-icon-button-root`.

Curious what others think! This could also help prevent any overlap with likely css classes customers already use.

stacysun-cb: Thanks @hcopp ! My only concern with adding prefixes and suffixes is the increase in HTML size and the potential performance impact. That said, I agree that a small prefix like cds is still valuable for avoiding name collisions and making components easy to target. I’d prefer something clean and minimal, e.g. `cds-IconButton`.

Additionally, I believe we want to avoid adding static class names to all subcomponents, for the same HTML size and performance concerns mentioned above.

hcopp: That's a fair point. We could still do the subcomponents only for the non primitive / more complex components. And we could apply that logic for our styles/classNames props.

If we don't include this path for the more complex components then it would limit use cases and force teams to use the classNames and styles props while other libraries support this third option which is very powerful to style all instances of components at once. I checked mui, ant, chakra, and mantine and they all include it.

We could do .cds-Button, .cds-Icon, etc and then .cds-Stepper-root, .cds-Stepper-step, .cds-Select-root, .cds-Select-control, etc

PR #266, opened 2025-12-19
fix: addressing misbehaving margin props for Banner

haoruikun-cb:
## Summary

In a previous PR (#209), we fixed a web `Banner` layout regression by moving from `flexGrow`-based sizing to explicit `width`/`height`. That change was necessary to resolve a production issue and also cleaned up a few edge-case layout glitches (see the original PR for details).

However, after the change, the `margin*` style props started behaving unexpectedly. CDS `margin*` props are implemented as **negative margins** to create a “bleed” effect. When the wrapper is sized with `width="100%"`, a negative horizontal margin simply **shifts** the Banner left/right while keeping its width at 100%, which *removes the intended bleed*. The visual result is that the Banner appears misaligned or clipped instead of extending past the container.

To restore the expected behavior, we now compute and apply an explicit wrapper width that accounts for negative margins (e.g. `calc(100% + <startBleed> + <endBleed>)`). This makes the Banner **span the full container and bleed past it** by the negative margin amount, and it correctly tracks responsive margin overrides at each breakpoint.

The earlier fix switched the Banner wrapper from `flexGrow` sizing to explicit `width`, but our `margin*` props are implemented as negative margins. With a fixed width="100%", negative horizontal margins only shift the element instead of expanding its visual footprint, so the intended “bleed” past the container was lost. This made margin-based layouts look misaligned or clipped.

## UI changes

Web Old

https://github.com/user-attachments/assets/176211ca-f1e6-46a8-b09c-e706dcda2e93

Web New 

https://github.com/user-attachments/assets/bfdeb8cb-17b8-40c5-927e-d690355193a6

https://github.com/user-attachments/assets/ceae5705-b4f0-470c-be03-4f1a496846e6

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/web/src/banner/__stories__/Banner.stories.tsx
+++ b/packages/web/src/banner/__stories__/Banner.stories.tsx
@@ -36,7 +36,7 @@ const exampleProps: ExampleProps = {
 
 const examplePropsWithMargin: ExampleProps = {
   ...exampleProps,
-  marginX: { phone: -2, desktop: 0 },
+  marginX: -2,
   children: shortMessage,
 };
 


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
🟡 See below
</td>
<td></td>
</tr>
</table>

<h2>
🟡
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
🟡
<code>0/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/banner/Banner.tsx
```diff
@@ -37,6 +45,119 @@ export const contentResponsiveConfig: ResponsiveProps<StaticStyleProps>['flexDir
...
+  computeForBreakpoint('desktop');
+
+  const hasResponsiveOverrides =
+    typeof widthByBreakpoint.phone !== 'undefined' ||
+    typeof widthByBreakpoint.tablet !== 'undefined' ||
+    typeof widthByBreakpoint.desktop !== 'undefined';
+
+  return hasResponsiveOverrides ? widthByBreakpoint : baseWidth;
+};
+
```
haoruikun-cb, line 160: Debate question: is it worth introducing complexity like this for addressing this specific use case?

packages/web/src/banner/Banner.tsx
```diff
@@ -37,6 +45,119 @@ export const contentResponsiveConfig: ResponsiveProps<StaticStyleProps>['flexDir
...
+const computeBleedWidth = (
+  effectiveMarginX: string | number | undefined,
+  effectiveMarginStart: string | number | undefined,
+  effectiveMarginEnd: string | number | undefined,
+) => {
+  // Prefer explicit start/end over marginX (marginX sets both sides).
+  const startSpaceToken =
+    normalizeNegativeSpace(effectiveMarginStart) ?? normalizeNegativeSpace(effectiveMarginX);
+  const endSpaceToken =
+    normalizeNegativeSpace(effectiveMarginEnd) ?? normalizeNegativeSpace(effectiveMarginX);
```
maximo-macchi-cb, line 66: Do other components prioritize `marginStart` / `marginEnd` styling over `marginX`? We should keep the styling API consistent with other components since this is a common styling prop

packages/web/src/banner/Banner.tsx
```diff
@@ -37,6 +45,119 @@ export const contentResponsiveConfig: ResponsiveProps<StaticStyleProps>['flexDir
...
 } as const;
 
+const normalizeNegativeSpace = (value: string | number | undefined) => {
+  if (value == null) return undefined;
+  const str = String(value);
+  if (!str.startsWith('-')) return undefined;
+  return str.slice(1);
+};
+
+const spaceVar = (spaceToken: string) => `var(--space-${spaceToken.replace('.', '_')})`;
```
maximo-macchi-cb, line 55: This might be better utilized inside the component and with the `useTheme` hook. We could avoid these normalization methods and just append `'px'` to the theme token value

packages/web/src/banner/Banner.tsx
```diff
@@ -201,7 +327,7 @@ export const Banner = memo(
       const content = (
         <Box
           position="relative"
-          width="100%"
+          width={showDismiss ? '100%' : wrapperWidth}
```
maximo-macchi-cb, line 330: If using `width="100%"` is the root cause for these styling issues, could it be removed? There could be another CSS solution which has the Banner's width grow without using flex. The PR you linked that introduced this regression was addressing an issue with the Banner's height. It would be better to address the root cause of this issue. This solution does seem a bit over-engineered. 

Could something like this work?

```suggestion
          // Replace with margin props based on order priority
          style={{ width: `calc(100% + ${marginX}px` }}
```

packages/web/src/banner/Banner.tsx
```diff
@@ -37,6 +45,119 @@ export const contentResponsiveConfig: ResponsiveProps<StaticStyleProps>['flexDir
...
+  const baseWidth = computeBleedWidth(respMarginX.base, respMarginStart.base, respMarginEnd.base);
+
+  const widthByBreakpoint: {
+    base: string;
+    phone?: string;
+    tablet?: string;
+    desktop?: string;
+  } = { base: baseWidth };
+
+  const computeForBreakpoint = (breakpoint: 'phone' | 'tablet' | 'desktop') => {
```
maximo-macchi-cb, line 131: Let's use the `DeviceBreakpoint` type from `src/styles/media.ts`

```suggestion
  const computeForBreakpoint = (breakpoint: Pick<DeviceBreakpoint, 'phone' | 'tablet' | 'desktop') => {
```

packages/web/src/banner/Banner.tsx
haoruikun-cb: Totally fair point on the complexity here. The reason we moved away from flex/grow is that in some edge cases the Banner ended up way off (example below; more in the prior PR). 
Image
In that PR, Stacy and I tried several approaches, and through trial/error we found `width="100%"` + `height="fit-content"` was the only combo that addressed all edge cases.

Your suggested approach (`style={{ width: calc(100% + ${marginX}px) }}`) also isn’t sufficient on its own because we support multiple margin props (start/end/X) and they’re `ResponsiveProp`s, so we need to resolve the correct breakpoint-specific values.

That said, I agree this is a lot of complexity to make one prop work — and it’s not frequently used. I was hesitant about this change and want to open this up for debate.

Possible alternatives:
1. **Don’t fix the margin behavior in Banner**. Instead, document a recommended workaround: wrap Banner in a `Box` with negative margin (as @hcopp [suggested](https://coinbase.slack.com/archives/C05H922EYP7/p1766097260734119?thread_ts=1766090660.647289&cid=C05H922EYP7) in #ask-cds).
2. **Auto-wrap when margin props are passed**. Detect margin props and insert an extra `Box` wrapper to apply the bleed. This avoids the width calc logic but introduces extra wrappers conditionally. It's still not ideal but reduced a lot of complexity. 
```return hasMarginProps ? <Box {...marginStyles}>{bannerBody}</Box> : bannerBody;```

Let me know which direction you all prefer.

haoruikun-cb: Hunter and I just had a quick Huddle and found a cleaner fix. I'll update the PR in a moment

PR #302, opened 2026-01-13
feat: chart legend

hcopp:
## What changed? Why?

This PR adds a new chart component, Legend.

It also adds support for hideBeaconLabels prop to Scrubber and BarComponent to BarChart.

### Web

image

image

image

image

image

image

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web-visualization/src/chart/legend/DefaultLegendEntry.tsx
```diff
@@ -0,0 +1,148 @@
...
+            style={styles?.shape}
+          />
+        </Box>
+        {typeof label === 'string' ? (
+          <Text className={classNames?.label} font="label1" style={styles?.label}>
+            {label}
+          </Text>
+        ) : (
+          label
+        )}
```
hcopp, line 52: I am waiting to hear back from design regarding if we should add aria-hidden to the default legend item or not, since users may expect to only hear the name of each series when scrubbing.

apps/docs/docs/components/graphs/LineChart/_webExamples.mdx
```diff
@@ -522,6 +506,7 @@ function Transitions() {
           },
         ]}
         yAxis={{ domain: { min: -domainLimit, max: domainLimit } }}
+        aria-hidden="true"
```
hcopp, line 509: This example and above were interfering with screen reader since they automatically update.

apps/mobile-app/scripts/utils/routes.mjs
```diff
@@ -170,6 +170,11 @@ export const routes = [
     getComponent: () =>
       require('@coinbase/cds-mobile/collapsible/__stories__/Collapsible.stories').default,
   },
+  {
+    key: 'Combobox',
+    getComponent: () =>
+      require('@coinbase/cds-mobile/alpha/combobox/__stories__/Combobox.stories').default,
+  },
```
hcopp, line 177: This was added when running the sync script

packages/mobile-visualization/src/chart/bar/Bar.tsx
```diff
@@ -146,6 +151,7 @@ export const Bar = memo<BarProps>(
         originY={effectiveOriginY}
         roundBottom={roundBottom}
         roundTop={roundTop}
+        seriesId={seriesId}
```
hcopp, line 154: This provides support for custom bar components

apps/docs/docs/components/graphs/Scrubber/_webExamples.mdx
```diff
@@ -365,6 +365,38 @@ function CustomBeaconLabel() {
 }
 ```
 
+You can use `hideBeaconLabels` to hide beacon labels, while still being able to provide a label for a series.
```
hcopp, line 358: This is a nice benefit of legend, teams can still label lines without the beacon labels

hcopp:
Tested this with design & confirmed accessibility, made a few changes. It is ready for review again :)

packages/mobile-visualization/src/chart/bar/BarChart.tsx
```diff
@@ -35,8 +30,9 @@ export type BarChartBaseProps = Omit<CartesianChartBaseProps, 'xAxis' | 'yAxis'
   > & {
     /**
      * Configuration objects that define how to visualize the data.
+     * Each series can optionally have its own BarComponent.
```
cb-ekuersch: nit: have => define

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  legendItem: {
+    alignItems: 'center',
+  },
+});
+
+export type LegendItemBaseProps = Omit<HStackProps, 'children' | 'color'> &
+  SharedProps & {
```
cb-ekuersch: nit: full types before Omits

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+    height: 24,
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  legendItem: {
+    alignItems: 'center',
+  },
+});
+
+export type LegendItemBaseProps = Omit<HStackProps, 'children' | 'color'> &
```
cb-ekuersch: nit: i think LegendEntry is more accurate

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Shape of the series.
+     *
+     */
+    shape?: LegendShape;
+    /**
+     * Custom component to render the legend shape.
+     * @default DefaultLegendShape
+     */
+    ShapeComponent?: LegendShapeComponent;
```
cb-ekuersch: nit: i think the shapes/colors are typically referred to as _keys_

packages/mobile-visualization/src/chart/legend/Legend.tsx
```diff
@@ -0,0 +1,126 @@
...
+import type { StyleProp, View, ViewStyle } from 'react-native';
+import type { SharedProps } from '@coinbase/cds-common/types';
+import { Box, type BoxProps } from '@coinbase/cds-mobile/layout';
+
+import { useCartesianChartContext } from '../ChartProvider';
+
+import { DefaultLegendItem, type LegendItemComponent } from './DefaultLegendItem';
+import type { LegendShapeComponent } from './DefaultLegendShape';
+
+export type LegendBaseProps = SharedProps & {
```
cb-ekuersch: nit: should this extend BoxBaseProps then?

packages/mobile-visualization/src/chart/legend/Legend.tsx
```diff
@@ -0,0 +1,126 @@
...
+   */
+  ShapeComponent?: LegendShapeComponent;
+  /**
+   * Accessibility label for the legend group.
+   * @default 'Legend'
+   */
+  accessibilityLabel?: string;
+};
+
+export type LegendProps = BoxProps &
```
cb-ekuersch: is there any value to extending BoxProps if we extend the Base box props above?

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+  ...props
+}) {
+  return (
+    <HStack
+      gap={gap}
+      style={[styles.legendItem, style, stylesProp?.root]}
+      testID={testID}
+      {...props}
+    >
+      <Box style={[styles.shapeWrapper, stylesProp?.shapeWrapper]}>
```
cb-ekuersch: nit: can we remove this extra element?

packages/mobile-visualization/src/chart/CartesianChart.tsx
```diff
@@ -64,6 +66,23 @@ export type CartesianChartBaseProps = Omit<BoxBaseProps, 'fontFamily'> &
      * Inset around the entire chart (outside the axes).
      */
     inset?: number | Partial<ChartInset>;
+    /**
+     * Whether to show the legend or a custom legend element.
+     * - `true` renders the default Legend component
+     * - A React element renders that element as the legend
+     * - `false` or omitted hides the legend
+     */
+    legend?: boolean | React.ReactElement<LegendProps>;
```
cb-ekuersch: traditionally we have accepted React components instead of JSX elements - what is the rationale here?

packages/mobile-visualization/src/chart/CartesianChart.tsx
```diff
@@ -64,6 +66,23 @@ export type CartesianChartBaseProps = Omit<BoxBaseProps, 'fontFamily'> &
...
+    /**
+     * Position of the legend relative to the chart.
+     * @default 'bottom'
+     */
+    legendPosition?: LegendPosition;
+    /**
+     * Accessibility label for the legend group.
+     * @default 'Legend'
+     */
+    legendAccessibilityLabel?: string;
```
cb-ekuersch, line 85: when and how do we determine what props we should surface on the parent component? One could argue that the label can be set by providing their own variaiton of the default legend component if the really needed to

packages/mobile-visualization/src/chart/CartesianChart.tsx
```diff
@@ -429,26 +451,62 @@ export const CartesianChart = memo(
...
+              >
+                {(legendPosition === 'top' || legendPosition === 'left') && legendElement}
+                <Box collapsable={collapsable} onLayout={onContainerLayout} style={{ flex: 1 }}>
+                  <ChartCanvas style={styles?.chart}>{children}</ChartCanvas>
+                </Box>
+                {(legendPosition === 'bottom' || legendPosition === 'right') && legendElement}
+              </Box>
+            ) : (
+              <Box
+                ref={ref}
```
cb-ekuersch: nit: you can capture the props to Box in a typed object literal and spread onto the two box elements

packages/web-visualization/src/chart/bar/BarStack.tsx
```diff
@@ -11,14 +11,24 @@ import { DefaultBarStack } from './DefaultBarStack';
...
+
 export type BarStackBaseProps = Pick<
   BarProps,
   'BarComponent' | 'fillOpacity' | 'stroke' | 'strokeWidth' | 'borderRadius'
 > & {
   /**
    * Array of series configurations that belong to this stack.
    */
-  series: Series[];
+  series: BarSeries[];
```
cb-ekuersch, line 31: do we need a cartesian series now?

packages/mobile-visualization/src/chart/CartesianChart.tsx
hcopp: I followed adding this based on Select Alpha's controlAccessibilityLabel. In my testing with design, they expressed the desire for this to be set when using a legend and as such not including this on the base component would require users to always pass in a react node to be compliant.

packages/web-visualization/src/chart/bar/BarStack.tsx
hcopp: I currently have this in our polar charts PR, since I realized then we would need PolarSeries and CartesianSeries. https://github.com/coinbase/cds/blob/hunter/polar-charts-2/packages/web-visualization/src/chart/utils/chart.ts#L38

Not sure if there is a benefit of adding it now vs then, but let me know what you think.

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
hcopp: Thanks for the suggestion. I am a little nervous to use 'key' since key and legend feel synonymous to me (and from what I see online).

Recharts [uses icon](https://recharts.github.io/en-US/api/Legend/#iconSize) for the naming, Material [uses mark](https://mui.com/x/react-charts/legend/#change-mark-shape).

I wasn't quite sure what to use for this, I see design uses 'item' for the shape + label combo, and then instead of shape they use 'media' ([link to figma](https://www.figma.com/design/zBWhJUcHCr5m8y1LULmPek/Pie-Donut-Charts?node-id=10647-2912&m=dev)).

 Here are a few sites I saw
- https://study.com/learn/lesson/map-key-symbols-examples.html
- https://english.stackexchange.com/questions/11878/legend-or-key
- https://www.ssc.education.ed.ac.uk/BSL/geography/keyd.html

packages/mobile-visualization/src/chart/legend/Legend.tsx
hcopp: On mobile BoxProps also adds generic React Native ViewProps https://github.com/coinbase/cds/blob/master/packages/mobile/src/layout/Box.tsx#L41. Web is similar, you get the polymorphic props for the element with BoxProps, you don't get those with just BoxBaseProps.

packages/mobile-visualization/src/chart/CartesianChart.tsx
hcopp: Good catch. Initially I was accepting just a boolean here, and then I saw the need to customize legend without wanting to add too many extra fields to CartesianChart.

I then for a bit was cloning the component passed in to add an additional field but realized this wasn't a great practice so I updated this.

Now it will be `boolean | React.ReactNode`. However I agree that this still might not be correct. Across the app we accept ReactNode in many places.

SlideButton accepts startUncheckedNode and endCheckedNode  and it also accepts SlideButtonHandleComponent and SlideButtonBackgroundComponent.

Select accepts ReactNode for label and also accepts SelectControlComponent.

In these cases we have been accepting the components in cases where we need to pass props, however for here we don't need to since the user can use context. But in the cases where we accept a reactnode the requests for the user are much simpler, maybe an Icon or Text component. Neither of these cases fully match Legend, this feel a little more uncharted.

If we switch to component we could support `showLegend` and `LegendComponent`. If we keep reactnode we can do `legend` and keep with boolean or reactnode.

packages/mobile-visualization/src/chart/legend/Legend.tsx
hcopp: Good catch and thanks for reviewing all of this in such depth!

packages/mobile-visualization/src/chart/bar/BarChart.tsx
hcopp: Thanks for reviewing this PR!

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
hcopp: Good thinking

packages/mobile-visualization/src/chart/legend/DefaultLegendItem.tsx
cb-ekuersch: Thanks for the resources! I trust your judgement!

packages/web-visualization/src/chart/bar/BarStack.tsx
cb-ekuersch: Fine to keep it where it is!

packages/mobile-visualization/src/chart/CartesianChart.tsx
cb-ekuersch: @hcopp so the user would need to render something that consumes the chart context under the hood? It almost feels simpler to support the LegendComponent and pass props that way. In that case I would suggest inverting the show to hide: `hideLegend`

Again, I trust your judgement on this. I think what rang the bell for me was `LegendProps` - in its current form the customization already forces them to write a component that mirrors our legend interface which could in theory just be passed as a custom component

hcopp: The prop is now `legend?: boolean | React.ReactNode` so we don't need to worry about LegendProps, that was leftover from earlier testing of mine when I was going to see if there was a benefit to cloning the node, and I have dropped that logic entirely.

Just spoke with Adrien. I think this current path is the simplest given the problem doesn't fully sit on either side of our traditional sub/default components or react node.

Someone can make their own Legend fully though and pass in via ReactNode now though!

PR #325, opened 2026-01-22
fix: Make elevation a prop for Switch

sverg-cb:
## What changed? Why?
- Expose `elevation` as a prop for both Web (needed) and Mobile (parity) versions of `Switch`

- User reported as shadow around the thumb for Switch and suggested the shadow should be optional

## UI changes

|  | Screenshot 2026-01-26 at 10 57 04 AM  |

| Screenshot 2026-01-26 at 10 37 51 AM | Screenshot 2026-01-26 at 10 36 32 AM Screenshot 2026-01-26 at 10 37 09 AM |

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/web/src/controls/Radio.tsx
+++ b/packages/web/src/controls/Radio.tsx
@@ -81,7 +81,14 @@ const RadioWithRef = forwardRef(function RadioWithRef<T extends string>(
   });
 
   return (
-    <Control ref={ref} checked={checked} elevation={elevation} label={children} type="radio" {...props}>
+    <Control
+      ref={ref}
+      checked={checked}
+      elevation={elevation}
+      label={children}
+      type="radio"
+      {...props}
+    >
       <Box
         alignItems="center"
         background={background}


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/controls/Switch.tsx
```diff
@@ -101,7 +106,7 @@ const SwitchWithRef = forwardRef<HTMLInputElement, SwitchProps>(function SwitchW
           borderRadius={1000}
           className={thumbCss}
           data-testid="switch-thumb"
-          elevation={1}
+          elevation={elevation}
```
hcopp, line 105: > User reported as shadow around the thumb for Switch and suggested the shadow should be optional

Did we confirm with design this was a bug? I don't see it in design but changing default behavior is not ideal. If they are wanting us to remove it then we should probably note that carefully in the web changelog that we are fixing a bug where there was an elevation on the switch.

packages/mobile/src/controls/Switch.tsx
```diff
@@ -90,12 +102,16 @@ const SwitchIcon = ({
...
+  { children, elevation = 0, ...props }: SwitchProps<T>,
   ref: React.ForwardedRef<View>,
 ) {
   const theme = useTheme();
   const { switchHeight } = theme.controlSize;
 
+  const renderSwitchIcon = (iconProps: ControlIconProps) => (
+    <SwitchIcon {...iconProps} elevation={elevation} />
+  );
+
```
hcopp: This feels off. Do you think we could try adding elevation to Control?

packages/web/src/controls/Switch.tsx
```diff
@@ -60,6 +64,7 @@ const SwitchWithRef = forwardRef<HTMLInputElement, SwitchProps>(function SwitchW
     children,
     checked,
     disabled,
+    elevation = 0,
```
hcopp: nit: can you remove `= 0`? I try to not include falsy defaults (such as false or 0). Same for the jsdocs

packages/web/src/controls/Switch.tsx
sverg-cb: Confirmed w/ Mandy in the original [Slack thread](https://coinbase.slack.com/archives/C05H922EYP7/p1768500423801769) that Switch should not have a drop shadow (by default) but allowed to be configurable

packages/mobile/src/controls/Switch.tsx
sverg-cb: Excellent callout, will do

packages/mobile/src/controls/Switch.tsx
```diff
@@ -81,6 +82,7 @@ const SwitchIcon = ({
         borderRadius={borderRadius}
         borderWidth={100}
         disabled={disabled}
+        elevation={elevation}
```
hcopp, line 85: I just checked and I think we should do this with Checkbox and Radio on mobile (I don't think we should for CheckboxCell and RadioCell though since the elevation there should be applied to the overall cell).

Or we could omit the prop. I think either case is better than Checkbox and Radio accepting the prop but not using it (current setup) or us having to recreate the component here in Switch as we were before).

apps/docs/docs/components/inputs/Switch/_mobileExamples.mdx
```diff
@@ -47,3 +47,19 @@ function AdvancedCustomColors() {
   );
 }
 ```
+
+### Elevation
+
+You can add a drop shadow to the Switch thumb using the `elevation` prop. By default, the Switch has no elevation.
+
+```jsx live
```
hcopp: nit: drop 'live' for mobile

packages/web/src/controls/Control.tsx
```diff
@@ -62,7 +62,7 @@ export type ControlBaseProps<T extends string> = FilteredHTMLAttributes<
   Partial<
     Pick<
       InteractableBaseProps,
-      'background' | 'borderColor' | 'borderRadius' | 'borderWidth' | 'color'
+      'background' | 'borderColor' | 'borderRadius' | 'borderWidth' | 'color' | 'elevation'
```
hcopp, line 65: Thanks for the update. Mobile now looks great.

On web though, we are now supporting elevation on Control but the components built on control such as Radio, Switch, and Checkbox aren't passing elevation into Control, they are placing it on their own components.

Web and mobile already have some inconsistencies here with Control. I think it could be fine to accept the prop on mobile control and on web leave it be just for Switch.

Or you could do what might be the more consistent/better in long run by checking if we can change the behavior of Switch, Radio, and Checkbox on web to pass in elevation to Control and see if it changes visually. If it looks the same on switch when elevation={1} no matter where we end up using the value (on Control or on the MotionBox then that is fine.

Sorry for any confusion!

packages/web/src/controls/Switch.tsx
hcopp, line 86: Did you try this with Switch?

sverg-cb: Screenshot 2026-01-27 at 11 54 20 AM

Looking good in latest local storybook

hcopp: Sorry I meant moving elevation over to Control. For the other two we set elevation on Control but here we set it on the Box.

Or confirming on Checkbox and Control that it looks fine and we accept on Switch that it is different.

hcopp: Actually I see now that this is specifically for the slider, not the component as a whole. I think if we can confirm the other two look fine as is with the elevation on Control then we should be good here!

sverg-cb: On Switch, the elevation is over the thumb specifically. Since there are 2 interactable pieces on Switch, I thought it made more sense for the shadow to follow the thumb instead of the entire track. Moving the prop to Control for Switch makes a shadow around entire switch component instead of only the piece that moves. Checkbox makes sense to have in `Control` since there is only 1 interactable element

lol i wish i'd seen your reply before I uploaded this 😅 

| Example | Story |
|--------|--------|
| Switch (shadow on thumb) | Screenshot 2026-01-27 at 12 22 45 PM Screenshot 2026-01-27 at 12 21 56 PM |
| Switch (shadow on track) | Screenshot 2026-01-27 at 12 20 54 PM  |
| Checkbox | Screenshot 2026-01-27 at 12 20 22 PM  |

hcopp: Gotcha, this looks fine then!

PR #352, opened 2026-01-30
Harryhao/cds 1513 consolidate listcell border props

haoruikun-cb:
## What changed? Why?

Updated Cell border handling so all border-related props are applied to the same inner container as `borderRadius` (web + mobile), and added focused border customization stories for both platforms. This resolves a customer-reported issue where borders were hard to configure because `bordered` applied to the outer Box while `borderRadius` applied to the inner Pressable, causing inconsistent results (see conversation: https://coinbase.slack.com/archives/C05H922EYP7/p1768246943588929).

Refactored Interactable to accept responsive `borderColor` and added coverage for breakpoint-based resolution. This improves Interactable/Pressable parity with other responsive props and makes border styling consistent across breakpoints.
## UI changes

| old screenshot | image |

| old screenshot | image |

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/web/src/system/Interactable.tsx
+++ b/packages/web/src/system/Interactable.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useContext, useMemo } from 'react';
+import React, { forwardRef, useMemo } from 'react';
 import { getBlendedColor } from '@coinbase/cds-common/color/getBlendedColor';
 import type { ThemeVars } from '@coinbase/cds-common/core/theme';
 import {
@@ -12,9 +12,9 @@ import { css } from '@linaria/core';
 import type { Polymorphic } from '../core/polymorphism';
 import type { Theme } from '../core/theme';
 import { cx } from '../cx';
+import { useResolveResponsiveProp } from '../hooks/useResolveResponsiveProp';
 import { useTheme } from '../hooks/useTheme';
 import { Box, type BoxBaseProps } from '../layout/Box';
-import { resolveResponsiveProp } from '../utils/responsive';
 
 import {
   interactableBackground,
@@ -28,7 +28,6 @@ import {
   interactablePressedBorderColor,
   interactablePressedOpacity,
 } from './interactableCSSProperties';
-import { MediaQueryContext } from './MediaQueryProvider';
 
 const COMPONENT_STATIC_CLASSNAME = 'cds-Interactable';
 
@@ -223,11 +222,7 @@ export const Interactable: InteractableComponent = forwardRef<
   ) => {
     const Component = (as ?? interactableDefaultElement) satisfies React.ElementType;
     const theme = useTheme();
-    const mediaQueryContext = useContext(MediaQueryContext);
-    const resolvedBorderColor = useMemo(
-      () => resolveResponsiveProp(borderColor, mediaQueryContext?.getSnapshot),
-      [borderColor, mediaQueryContext?.getSnapshot],
-    );
+    const resolvedBorderColor = useResolveResponsiveProp(borderColor);
 
     const interactableStyle = useMemo(
       () => ({
--- a/packages/web/src/cells/__stories__/ListCell.stories.tsx
+++ b/packages/web/src/cells/__stories__/ListCell.stories.tsx
@@ -921,7 +921,11 @@ const BorderCustomization = () => {
       />
       <ListCell
         bordered
-        borderColor="accentBoldPurple"
+        borderColor={{
+          phone: 'fgPositive',
+          tablet: 'fgNegative',
+          desktop: 'accentBoldPurple',
+        }}
         description="Custom border color"
         onClick={onClickConsole}
         spacingVariant={spacingVariant}
--- a/packages/web/src/hooks/__tests__/useResolveResponsiveProp.test.tsx
+++ b/packages/web/src/hooks/__tests__/useResolveResponsiveProp.test.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+import { renderHook } from '@testing-library/react';
+
+import { media } from '../../styles/media';
+import { MediaQueryContext } from '../../system/MediaQueryProvider';
+import { useResolveResponsiveProp } from '../useResolveResponsiveProp';
+
+const createMediaContext = (width: number) => ({
+  subscribe: () => () => {},
+  getServerSnapshot: (query: string) => {
+    if (query === media.phone) return width <= 767;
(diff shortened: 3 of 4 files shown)


packages/mobile/src/cells/__stories__/ListCell.stories.tsx
```diff
@@ -794,6 +794,146 @@ const WithHelperText = () => (
...
+
+  return (
+    <VStack gap={2} padding={2}>
+      <Switch
+        checked={isCondensed}
+        onChange={(_, nextChecked) => setIsCondensed(Boolean(nextChecked))}
+      >
+        Spacing variant: {spacingVariant}
+      </Switch>
+      <ListCell
```
cb-ekuersch, line 809: nitL are this many examples necessary? Try to limit stories to a few key examples. Keep in mind the primary use case is visual regression testing and visual documentation/interaction.

packages/mobile/src/cells/Cell.tsx
```diff
@@ -143,9 +160,52 @@ export const Cell = memo(function Cell({
 
   const { marginX: innerSpacingMarginX, ...innerSpacingWithoutMarginX } = innerSpacing;
 
+  const borderProps = useMemo(
```
cb-ekuersch, line 166: please leave a comment:
- explaining why we need to collect the border props 
- a link to the linear issue we created to remove the outer Cell element (i believe we have one, if not lets create a breaking change cds-v10 issue for it)

packages/web/src/system/Interactable.tsx
```diff
@@ -27,6 +29,7 @@ import {
   interactablePressedBorderColor,
   interactablePressedOpacity,
 } from './interactableCSSProperties';
+import { MediaQueryContext } from './MediaQueryProvider';
```
cb-ekuersch: this is technically a breaking change. Before Interactible didn't have to be used inside the MediaQueryProvider (even tho it probably is in most cases already)

packages/web/src/system/Interactable.tsx
```diff
@@ -106,6 +109,24 @@ const transparentWhileInactiveCss = css`
   }
 `;
 
+const isResponsiveValue = <T,>(value: ResponsiveProp<T>): value is ResponsiveValue<T> =>
```
cb-ekuersch: nit: lets put this in a web util. Maybe if we already have a `responsive` module or something similar it can live there

packages/web/src/system/Interactable.tsx
```diff
@@ -163,8 +184,6 @@ export type InteractableBaseProps = Polymorphic.ExtendableProps<
     background?: ThemeVars.Color;
     /** Set element to block and expand to 100% width. */
     block?: boolean;
-    /** Border color of the element. */
-    borderColor?: ThemeVars.Color;
```
cb-ekuersch, line 167: because this comes from Box right?

packages/web/src/system/Interactable.tsx
```diff
@@ -223,18 +242,23 @@ export const Interactable: InteractableComponent = forwardRef<
   ) => {
     const Component = (as ?? interactableDefaultElement) satisfies React.ElementType;
     const theme = useTheme();
+    const mediaQueryContext = useContext(MediaQueryContext);
+    const resolvedBorderColor = useMemo(
```
cb-ekuersch: why do we need to do this? I thought responsive props resolve for us automatically?

packages/web/src/system/Interactable.tsx
haoruikun-cb: I think this should be an OK breaking change to make since Pressable are being used inside  `MediaQueryProvider` as well so customer technically don't need to change anything in most cases. But we need to make this thing clear in our v9 documentation so customer is aware of this.

packages/web/src/system/Interactable.tsx
haoruikun-cb: Yes that's correct.

packages/web/src/system/Interactable.tsx
haoruikun-cb: If the borderColor is being applied to `Box` directly, we shouldn't need to write this logic. However it's being applied to a `style` prop. It has a dedicated logic to compute what style is being applied to the Box based on the colors passed in:
image

packages/mobile/src/cells/__stories__/ListCell.stories.tsx
haoruikun-cb: Ok that make sense, I'll cut down the story here and only leave interesting/important examples

haoruikun-cb:
> please yarn install and verify your unit tests, lint and typecheck commands pass

There are some typecheck and lint issues coming from cds-v9 branch. I can draft a separate PR to fix that

packages/web/src/utils/responsive.ts
```diff
@@ -0,0 +1,47 @@
...
+ * media-query CSS variables.
+ *
+ * @param value - A scalar value or responsive object with base/phone/tablet/desktop keys
+ * @param getSnapshot - Function that returns whether a media query matches. Pass
+ *   MediaQueryContext.getSnapshot when used within MediaQueryProvider. Without it,
+ *   returns the first defined value (base ?? phone ?? tablet ?? desktop).
+ * @returns The resolved value for the current breakpoint, or the fallback when
+ *   getSnapshot is not provided
+ */
+export const resolveResponsiveProp = <T>(
```
cb-ekuersch: @haoruikun-cb  it might be more useful to make this a react hook, otherwise we have to always get the context before using this function

```
function useResolveResponsiveProp(value?: ResponsiveProp<T>) {
  const { getSnapshot } = useContext(MediaQueryContext);

  if (!value || !isResponsiveValue(value)) return value;
  const fallback = value.base ?? value.phone ?? value.tablet ?? value.desktop;
  if (!getSnapshot) return fallback;
  if (typeof value.phone !== 'undefined' && getSnapshot(media.phone)) return value.phone;
  if (typeof value.tablet !== 'undefined' && getSnapshot(media.tablet)) return value.tablet;
  if (typeof value.desktop !== 'undefined' && getSnapshot(media.desktop)) return value.desktop;
  return fallback;
}
```

haoruikun-cb: Yeah make sense

PR #566, opened 2026-03-30
feat: PopoverPanel + Dropdown deprecation

adrienzheng-cb:
### Popover (internal)
- Replace Popper.js positioning with @floating-ui/react-dom: useFloating, autoUpdate,
  offset (skid/gap), flip + shift + limitShift for fixed placements, and autoPlacement
  for placement values that start with "auto".
- Wrap the component in forwardRef and merge the consumer ref with the Floating UI
  reference ref (useMergeRefs).
- Add optional style and className on the subject (trigger) root.
### PopoverPanel / PopoverPanelContent
- Add PopoverPanel: trigger + floating or phone modal layout, content as node or
  (closePopover) => node, optional enableMobileModal, sizing (panelWidth, min/max
  width/height), showOverlay, contentPosition, ref with openPopover/closePopover.
- Add PopoverPanelContent: animated VStack surface for the panel body.
- Wire useComponentConfig for PopoverPanel and PopoverPanelContent; register both in
  core/componentConfig.ts.
- Export from packages/web overlays barrel and new package entry @coinbase/cds-web/popover
  (package.json exports).
- Add Storybook stories, unit tests, and Docusaurus (docgen, examples, props table,
  styles explorer, sidebar, webMetadata).
### useResponsivePanelMaxHeight (renamed from useResponsiveHeight in popover)
- New module overlays/popover/useResponsivePanelMaxHeight.ts (canonical hook for
  PopoverPanel).
- Rename the hook to useResponsivePanelMaxHeight; return value is panelMaxHeight
  (was dropdownHeight in the shared logic).
- Rename the measured bounds argument to panelBounds (was dropdownBounds) for panel
  sizing; keep the same viewport / calc(100vh - …) responsive behavior.
- Support ResponsiveProp<maxHeight> with resolveResponsiveMaxHeight aligned to
  phone/tablet/desktop breakpoints via useBreakpoints.
- Legacy packages/web/src/dropdown/useResponsiveHeight.ts remains for Dropdown only,
  with the original non-responsive maxHeight API; deprecate in favor of importing
  useResponsivePanelMaxHeight from @coinbase/cds-web/popover (removal v10) and add a
  short comment that new behavior lives under overlays/popover.
### Dropdown deprecation
- Deprecate Dropdown, DropdownContent, DropdownProps-related types, and docs
  Dropdown webMetadata warning: use PopoverPanel / PopoverPanelContent instead
  (expected removal v10).
### Other
- ModalWrapper: when hideOverlay is true, render a transparent fixed VStack that still
  receives overlay press so dismiss works without a visible Overlay.
- Jest: extend @floating-ui/react-dom mock with limitShift.
### Breaking / migration notes for consumers
- Popover: possible subtle positioning and DOM structure differences vs Popper; new
  ref and subject style/className are additive.
- Dropdown: deprecated; migrate to PopoverPanel.
- Popover-adjacent hook: prefer useResponsivePanelMaxHeight from @coinbase/cds-web/popover;
  dropdown/useResponsiveHeight is legacy and deprecated.

## UI changes

[storybook](https://69cecc739d43077bbebe49ee--cds-storybook.netlify.app/?path=/story/components-overlay-popoverpanel--default)
[docs](https://69cecc9be413c679ee5d5f1e--cds-docs.netlify.app/components/overlay/PopoverPanel/?tab=examples)

https://github.com/user-attachments/assets/f0f8392d-b9b1-4f29-9365-f7b495e1c400

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-ekuersch (commented):
Before we can recommend switching to Popover we should at least:
1. audit Popover to make sure it can handle any/all use cases that Dropdown can. Any storybook stories from Dropdown should be able to be effortlessly migrated to Popover stories
2. add docsite pages for Popover

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>
<details>
<summary><h2>⚠️ Ignored Reviews (2)</h2></summary>
<table>
<tr>
<th>Reviewer</th>
<th>Reason</th>
</tr>
<tr>
<td>cb-ekuersch</td>
<td>Old or outdated review</td>
</tr>
<tr>
<td>cb-ekuersch</td>
<td>Review missing MFA</td>
</tr>
</table>
</details>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/overlays/popover/Popover.tsx
```diff
@@ -1,6 +1,17 @@
 /* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
-import React, { memo, useCallback, useMemo } from 'react';
+import React, { forwardRef, memo, useCallback, useMemo } from 'react';
 import { zIndex } from '@coinbase/cds-common/tokens/zIndex';
+import { useMergeRefs } from '@coinbase/cds-common/hooks/useMergeRefs';
```
adrienzheng-cb, line 3: will replace with mergeRefs after merging into v9

packages/web/src/overlays/popover/Popover.tsx
```diff
@@ -1,6 +1,17 @@
...
+import {
+  autoPlacement,
+  autoUpdate,
+  flip,
+  limitShift,
+  offset,
+  type Placement as FloatingPlacement,
+  shift,
+  useFloating,
+} from '@floating-ui/react-dom';
```
adrienzheng-cb, line 14: originally replacing with popper with floating UI in popover was done in v9 feature branch, but moved it here since we needed to have PopoverPanel before v9 and Popover is affected

adrienzheng-cb: non-breaking since all props are the same

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,363 @@
...
+   */
+  panelWidth?: React.CSSProperties['width'];
+  /** Minimum width of the panel as a percentage string or number converted to pixels. */
+  minPanelWidth?: React.CSSProperties['minWidth'];
+  /** Maximum width of the panel as a percentage string or number converted to pixels. */
+  maxPanelWidth?: React.CSSProperties['maxWidth'];
+  /** Can optionally pass a maxHeight.
+   * @default 300
+   */
+  maxPanelHeight?: React.CSSProperties['maxHeight'];
```
adrienzheng-cb: maxPanelWidth/Height instead of maxHeight to make it clear to which element dimension is set

packages/web/src/overlays/modal/ModalWrapper.tsx
```diff
@@ -106,13 +107,20 @@ export const ModalWrapper = memo(
...
               <>
-                {!hideOverlay && (
+                {!hideOverlay ? (
                   <Overlay
                     animated
                     className={cx(!dangerouslyDisableResponsiveness && modalOverlayResponsiveCss)}
                     onClick={!disableOverlayPress ? onOverlayPress : undefined}
                     testID="modal-overlay"
                   />
+                ) : (
```
adrienzheng-cb, line 117: adding this here so Modal without overlay can still be closed when clicking outside

packages/web/src/dropdown/useResponsiveHeight.ts
```diff
@@ -18,6 +18,11 @@ type UseResponsiveHeightParams = {
 
 const BOTTOM_GUTTER_SPACE: ThemeVars.Space = 2;
 
+// New behavior (ResponsiveProp maxHeight, breakpoint resolution) lives in `overlays/popover/useResponsiveHeight.ts`; this file is legacy for Dropdown only.
+/**
+ * @deprecated Import `useResponsiveHeight` from `@coinbase/cds-web/popover` instead. This will be removed in a future major release.
+ * @deprecationExpectedRemoval v10
+ */
```
adrienzheng-cb, line 24: added a replacement under popover directory. the replacement handle can handle maxHeight with responsive value

packages/web/src/overlays/popover/PopoverProps.ts
```diff
@@ -83,4 +83,12 @@ export type PopoverProps = PopoverBaseProps & {
   onBlur?: (event?: React.FocusEvent) => void;
   /** Callback fired when a mouse down event is fired on the subject */
   onMouseDown?: (event: React.MouseEvent) => void;
+  /**
```
adrienzheng-cb, line 88: added style can className support to Popover

packages/web/src/overlays/popover/__stories__/PopoverPanel.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+import { VStack } from '../../../layout/VStack';
+import { Text } from '../../../typography/Text';
+import { PortalProvider } from '../../PortalProvider';
+import { PopoverPanel, type PopoverPanelRef } from '../PopoverPanel';
+
+const meta: Meta<typeof PopoverPanel> = {
+  title: 'Components/Overlay/PopoverPanel',
+  component: PopoverPanel,
+  parameters: {
+    a11y: { test: 'off' },
```
maximo-macchi-cb: Why are a11y tests being disabled? I enabled them locally and in the Storybook UI and CLI all tests pass

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Width of the panel as a percentage string or number converted to pixels.
+     */
+    panelWidth?: PopoverPanelContentBaseProps['width'];
+    /** Minimum width of the panel as a percentage string or number converted to pixels. */
+    minPanelWidth?: PopoverPanelContentBaseProps['minWidth'];
+    /** Maximum width of the panel as a percentage string or number converted to pixels. */
+    maxPanelWidth?: PopoverPanelContentBaseProps['maxWidth'];
+    /** Height of the panel as a percentage string or number converted to pixels. */
+    panelHeight?: PopoverPanelContentBaseProps['height'];
```
maximo-macchi-cb: Small nit: since our other height / width props only are named as "height" and "width" I think it'd be easier for engineers to pick up on by keeping the naming convention consistent

```suggestion
    /**
     * Width of the panel as a percentage string or number converted to pixels.
     */
    width?: PopoverPanelContentBaseProps['width'];
    /** Minimum width of the panel as a percentage string or number converted to pixels. */
    minWidth?: PopoverPanelContentBaseProps['minWidth'];
    /** Maximum width of the panel as a percentage string or number converted to pixels. */
    maxWidth?: PopoverPanelContentBaseProps['maxWidth'];
    /** Height of the panel as a percentage string or number converted to pixels. */
    height?: PopoverPanelContentBaseProps['height'];
```

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+        event.stopPropagation();
+      }, []);
+      return (
+        // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+        <div
+          className={cx(block ? blockCss : triggerContainerCss, classNames?.triggerContainer)}
+          onBlur={onBlur}
+          onClick={disabled ? undefined : onOpen}
+          style={styles?.triggerContainer}
+        >
```
maximo-macchi-cb, line 192: Even though this is meant for mobile devices, this could still render in a web browser on a desktop where the browser's window width is small, right? Two suggestions for this:
1. An `onKeyDown` handler should be added here
2. A prop for `accessibilityRole` with default value of `"button"` should be added to the prop API and used here

```suggestion
        <div
          accessibilityRole={accessibilityRole}
          className={cx(block ? blockCss : triggerContainerCss, classNames?.triggerContainer)}
          onBlur={onBlur}
          onClick={disabled ? undefined : onOpen}
          // onKeyDown could handle the key event then call onOpen
          onKeyDown={disabled ? undefined : onKeyDown}
          style={styles?.triggerContainer}
        >
```

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+          {children}
+          <ModalWrapper
+            dangerouslyDisableResponsiveness
+            disablePortal={disablePortal}
+            hideOverlay={!showOverlay}
+            onClick={handleCaptureEvents}
+            onOverlayPress={onClose}
+            testID="popover-panel-modal"
```
maximo-macchi-cb: We should make the `testID` a prop with default of `"popover-panel-modal"` so customers can configure this as they'd like

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+            visible={visible}
+            {...controlledElementAccessibilityProps}
+          >
+            <FocusTrap
+              onEscPress={onClose}
+              respectNegativeTabIndex={respectNegativeTabIndex}
+              restoreFocusOnUnmount={restoreFocusOnUnmount}
+            >
```
maximo-macchi-cb, line 211: I think adding the following props from `FocusTrapBaseProps` would be beneficial for customers to configure:
- `disableAutoFocus`
- `includeTriggerInFocusTrap`
- `focusTabIndexElements`
- `autoFocusDelay`

packages/web/src/overlays/popover/PopoverPanelContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+    return (
+      <MotionVStack
+        ref={ref}
+        bordered
+        background="bg"
+        borderRadius={400}
+        className={cx(popoverPanelContentClassName, className)}
+        elevation={2}
+        minWidth={minWidth}
+        overflow="auto"
```
maximo-macchi-cb: Shouldn't these be configurable? At least `background`, `borderRadius`, and `overflow`

maximo-macchi-cb (commented):
Currently reviewed new PopoverPanel components in web. Will finish review tomorrow but saving here just in case.

packages/web/src/core/componentConfig.ts
```diff
@@ -154,6 +156,8 @@ export type ComponentConfig = {
   PageFooter?: ConfigResolver<PageFooterBaseProps>;
   PageHeader?: ConfigResolver<PageHeaderBaseProps>;
   Pagination?: ConfigResolver<PaginationBaseProps>;
+  PopoverPanel?: ConfigResolver<PopoverPanelBaseProps>;
+  PopoverPanelContent?: ConfigResolver<PopoverPanelContentBaseProps>;
```
cb-ekuersch: Customers never explicitly render a `PopoverPanelContent` component. It is an implementation detail of the `content` prop. I recommend we don't add "hidden" components to the component theme config

packages/web/src/overlays/popover/__stories__/PopoverPanel.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+        testID="popover-panel-default"
+      >
+        <Button>Open panel</Button>
+      </PopoverPanel>
+    </HStack>
+  );
+}
+
+export const Default: Story = {
+  render: () => <DefaultStory />,
```
cb-ekuersch: i think there is a shorter way to define these, take a moment to review: https://storybook.js.org/blog/storybook-csf3-is-here/

You can have a default export that defines a render and then export objects that change the props/args. Not sure if that setup would help you at all here

packages/web/src/overlays/popover/__tests__/PopoverPanel.test.tsx
```diff
@@ -0,0 +1,106 @@
+import { createRef } from 'react';
+import useMeasure from 'react-use-measure';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+
+import { MediaQueryProvider } from '../../../system/MediaQueryProvider';
+import { DefaultThemeProvider } from '../../../utils/test';
+import { PopoverPanel, type PopoverPanelRef } from '../PopoverPanel';
+
+jest.mock('react-use-measure');
```
cb-ekuersch, line 9: nit: i wonder if we should just centralize the mock of this if its always necessary to set up, similar to what we did for floating ui

packages/web/src/overlays/index.ts
```diff
@@ -10,6 +10,8 @@ export * from './modal/ModalFooter';
 export * from './modal/ModalHeader';
 export * from './overlay/Overlay';
 export * from './popover/Popover';
+export * from './popover/PopoverPanel';
+export * from './popover/PopoverPanelContent';
```
cb-ekuersch: do we really want to to encourage direct use by putting it in the barrel file? It seems like this component is more an implementation detail; a true sub-component

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+export type PopoverPanelRef = {
+  openPopover: () => void;
+  closePopover: () => void;
+};
+
+export type PopoverPanelRenderContent = (api: { closePopover: () => void }) => React.ReactNode;
+
+export type PopoverPanelBaseProps = Pick<
+  PopoverBaseProps,
+  | 'children'
```
cb-ekuersch: put children on props, not base props

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+  | 'showOverlay'
+  | 'contentPosition'
+  | 'block'
+  | 'disableTypeFocus'
+  | 'controlledElementAccessibilityProps'
+  | 'respectNegativeTabIndex'
+> &
+  SharedProps &
```
cb-ekuersch: nit: put full types first in set

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+    restoreFocusOnUnmount?: boolean;
+    /**
+     * Panel body, or a function that receives `closePopover` (helpfulwhen actions inside the panel should dismiss it).
+     */
+    content: React.ReactNode | PopoverPanelRenderContent;
+  };
+
+export type PopoverPanelProps = PopoverPanelBaseProps & {
+  style?: React.CSSProperties;
+  styles?: {
```
cb-ekuersch: use styles and classnames helper - maybe try the skill Hunter added to see how it performs

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+    triggerContainer?: string;
+  };
+};
+
+export type PopoverPanelInternalProps = Omit<PopoverPanelProps, 'content'> & {
+  content: React.ReactNode;
+  visible: boolean;
+};
+
+export const POPOVER_PANEL_MAX_HEIGHT = 300;
```
cb-ekuersch: it might make more sense to encapsulate this in the module for the max height hook

packages/web/src/overlays/popover/PopoverPanel.tsx
```diff
@@ -0,0 +1,380 @@
...
+const NOOP = () => {};
+
+const defaultPopoverContentPositionConfig: PopoverContentPositionConfig = {
+  gap: 0.5,
+  placement: 'bottom-start',
+};
+
+function usePopoverPanelImperativeHandle(
```
cb-ekuersch, line 121: nit: is this separate hook necessary?

cb-ekuersch (commented):
Looking good, but I have a couple of quesitons:
1. What are the long term plans for the two popover directories: src/overlays/popover and  src/popover?
2. Are we planning to keep supporting Popover on its own? I thought PopoverPanel was meant to completely replace it rather than compose it

packages/web/src/dropdown/Dropdown.tsx
```diff
@@ -218,6 +218,10 @@ const PopoverDropdown = memo(
   ),
 );
 
+/**
+ * @deprecated Use PopoverPanel instead. This will be removed in a future major release.
```
maximo-macchi-cb: Select should be mentioned here too. Dropdown was very similar to Select. So while PopoverPanel may satisfy some customer's needs, others may be wanting a Select-like component and Select / SelectChip would be better for that than PopoverPanel

packages/web/src/overlays/modal/ModalWrapper.tsx
maximo-macchi-cb: Would moving this to the parent `<Box>` be a better alternative? Or does that not work?

maximo-macchi-cb (commented):
Review fully done! Left a couple additional comments

packages/web/src/core/componentConfig.ts
adrienzheng-cb: got it. I'm gonna remove TooltipContent from the component theme config for consistency then.

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: I added the panel prefix on purpose so it is clear that they define dimensions of the popover panel not the popover trigger.

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: the panel already opens when pressing enter while focusing on the children at the moment without `onKeyDown` and since this component handles the open and close logic internally and does not take onKeyDown prop, we don't need to pass onKeydown here.

for accessibilityRole, the child is most likely a button already, if we add additional role to the trigger wrapper here, then we end up with nested buttons.

I will mention in the docs that they should use always interactive element as the child(trigger) of PopoverPanel

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: it is separate because we use it both in Mobile(Modal) and FloatingPopoverPanel sub components.

packages/web/src/dropdown/Dropdown.tsx
adrienzheng-cb: updated all deprecation messages to recommend Select for selectable use case

packages/web/src/overlays/modal/ModalWrapper.tsx
adrienzheng-cb: I tried that at first, but the problem with that was clicking inside the dialog would also close the modal since the parent Box wraps both the overlay and the dialog.

packages/web/src/overlays/popover/__stories__/PopoverPanel.stories.tsx
adrienzheng-cb: oh my bad. I re-enabled it

packages/web/src/overlays/popover/__stories__/PopoverPanel.stories.tsx
adrienzheng-cb: removed this redundant code
exporting the DefaultStory component would do it.

packages/web/src/overlays/popover/__tests__/PopoverPanel.test.tsx
adrienzheng-cb: good suggestion! I will address this in a separate pr since we have a lot of tests that mock this hook.

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: fixed!

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: fixed!

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: fixed!

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: I ended up removing the max height hook. it is no longer necessary because floating-ui can already handle that for us.

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: added the testID prop but did not assign the default id because we are actually not using it and since this is a new component so no one else is selecting the component using this id.

packages/web/src/overlays/popover/PopoverPanel.tsx
adrienzheng-cb: done!

packages/web/src/overlays/popover/PopoverPanelContent.tsx
adrienzheng-cb: done!

packages/web/src/overlays/index.ts
adrienzheng-cb: good catch. I removed the PopoverPanelContent and made it strictly internal

adrienzheng-cb:
@cb-ekuersch 

> What are the long term plans for the two popover directories: src/overlays/popover and src/popover?

Good catch! I removed the src/popover directory. It was added by mistake. PopoverPanel and Popover are only exported from overlays now.

>  Are we planning to keep supporting Popover on its own? I thought PopoverPanel was meant to completely replace it rather than compose it

yes, we still need to support Popover on its own because I have seen consumers using this component in many places and implemented their own PopoverPanel/Dropdown logic on top of it.

cb-ekuersch (approved):

cb-ekuersch (approved):

PR #638, opened 2026-04-23
fix: configure dotted area to use mask

hcopp:
## What changed? Why?

This PR uses a shader for dotted area, which is more performant and fixes an issue which was causing crashes in v9 when the path would cross over itself.

This is already being used in retail.

Our dotted path had a custom set of calculations for dots within the area. When the path would cross over itself, it caused an invalid path svg to be made. Our new path doesn't have this issue, as it is like a regular area which has a mask. So a crossed over path doesn't have this issue.

Separately we are looking at the crossed over issue.

## UI changes

| image | image |

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/apps/test-expo/src/playground/ExamplesListScreen.tsx
+++ b/apps/test-expo/src/playground/ExamplesListScreen.tsx
@@ -53,14 +53,10 @@ export function ExamplesListScreen({ route }: ExamplesListScreenProps) {
     });
 
   return (
-    <Box
-      background="bg"
-      flexGrow={1}
-      style={{ paddingBottom: bottom }}
-      testID="mobile-playground-home-screen"
-    >
+    <Box background="bg" flexGrow={1} testID="mobile-playground-home-screen">
       <FlatList
         ItemSeparatorComponent={null}
+        contentContainerStyle={{ paddingBottom: bottom }}
         data={data}
         initialNumToRender={14}
         renderItem={renderItem}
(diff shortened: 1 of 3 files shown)


packages/mobile/src/visualizations/chart/utils/path.ts
```diff
@@ -355,6 +355,9 @@ export const getBarPath = (
  *   2  // 2px radius dots
  * );
  * ```
+ *
+ * @deprecated Prefer a shader for dotted areas instead. This will be removed in a future major release.
+ * @deprecationExpectedRemoval v10
```
hcopp, line 360: This is no longer needed

apps/test-expo/src/playground/ExamplesListScreen.tsx
```diff
@@ -53,14 +53,10 @@ export function ExamplesListScreen({ route }: ExamplesListScreenProps) {
...
-    <Box
-      background="bg"
-      flexGrow={1}
-      style={{ paddingBottom: bottom }}
-      testID="mobile-playground-home-screen"
-    >
+    <Box background="bg" flexGrow={1} testID="mobile-playground-home-screen">
       <FlatList
         ItemSeparatorComponent={null}
+        contentContainerStyle={{ paddingBottom: bottom }}
```
hcopp, line 59: This was causing an annoying issue where the text was cutoff. We still have padding, but now only at the bottom of the scroll view. When scrolling (and not at bottom) we don't have extra padding.

Before:




After:

haoruikun-cb (approved):

PR #644, opened 2026-04-27
fix: chart issues with v9 dependencies

hcopp:
## What changed? Why?

This PR fixes several issues with charts that were found in v9 branch

1. Scrubber label opacity was always at 1
2. Axes were getting added too late and caused the rest of the chart to animate when it shouldn't
3. Our ScrubberBeaconLabels were animating on initial render (this might have always been happening but not visible when animations are enabled)
4. Enter opacity for bar animations wasn't working

Listed out root causes in each file

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/mobile/src/visualizations/chart/CartesianChart.tsx
+++ b/packages/mobile/src/visualizations/chart/CartesianChart.tsx
@@ -1,9 +1,14 @@
-import React, { forwardRef, memo, useCallback, useMemo } from 'react';
-import { type StyleProp, type View, type ViewStyle } from 'react-native';
+import React, { forwardRef, memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import {
+  type LayoutChangeEvent,
+  type LayoutRectangle,
+  type StyleProp,
+  type View,
+  type ViewStyle,
+} from 'react-native';
 import type { Rect } from '@coinbase/cds-common/types';
 import { Canvas, Skia, type SkTypefaceFontProvider } from '@shopify/react-native-skia';
 
-import { useLayout } from '../../hooks/useLayout';
 import type { BoxBaseProps, BoxProps } from '../../layout';
 import { Box } from '../../layout';
 
@@ -202,7 +207,32 @@ export const CartesianChart = memo(
       },
       ref,
     ) => {
-      const [containerLayout, onContainerLayout] = useLayout();
+      const [containerLayout, setContainerLayout] = useState<Rect>({
+        width: 0,
+        height: 0,
+        x: 0,
+        y: 0,
+      });
+      const containerLayoutRafRef = useRef<number | null>(null);
+
+      const onContainerLayout = useCallback((event: LayoutChangeEvent) => {
+        const layout = event.nativeEvent.layout;
+        if (containerLayoutRafRef.current !== null) {
+          cancelAnimationFrame(containerLayoutRafRef.current);
+        }
+        containerLayoutRafRef.current = requestAnimationFrame(() => {
+          containerLayoutRafRef.current = null;
+          setContainerLayout(layout);
+        });
+      }, []);
+
+      useEffect(() => {
+        return () => {
+          if (containerLayoutRafRef.current !== null) {
+            cancelAnimationFrame(containerLayoutRafRef.current);
+          }
+        };
+      }, []);
 
       const chartWidth = containerLayout.width;
       const chartHeight = containerLayout.height;
--- a/packages/mobile/src/visualizations/chart/CartesianChart.tsx
+++ b/packages/mobile/src/visualizations/chart/CartesianChart.tsx
@@ -1,7 +1,6 @@
 import React, { forwardRef, memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import {
   type LayoutChangeEvent,
-  type LayoutRectangle,
   type StyleProp,
   type View,
   type ViewStyle,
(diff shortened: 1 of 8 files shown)


packages/mobile/src/visualizations/chart/Path.tsx
```diff
@@ -272,13 +266,11 @@ export const Path = memo<PathProps>((props) => {
...
     }
 
-    if (enterOpacityTransition === undefined || enterOpacityTransition === null) {
+    if (enterOpacityTransition == null) {
       enterOpacity.value = 1;
-      hasAnimatedEnterOpacity.current = true;
       return;
     }
 
-    hasAnimatedEnterOpacity.current = true;
```
hcopp, line 281: Looked at code examples in latest v4 docs to create a style that matches docs - https://docs.swmansion.com/react-native-reanimated/docs/animations/withSpring/.

Before, we were not seeing the chart fade in, I believe this happened  due to the useRef becoming true but other shared values being reset similar to strict mode.

packages/mobile/src/visualizations/chart/scrubber/Scrubber.tsx
```diff
@@ -434,7 +440,7 @@ export const Scrubber = memo(
       if (!isReady) return;
 
       return (
-        <Group opacity={scrubberOpacity}>
+        <Group layer={<Paint opacity={scrubberOpacity} />}>
```
hcopp, line 443: Followed https://github.com/Shopify/react-native-skia/issues/2210

packages/mobile/src/visualizations/chart/bar/__stories__/BarChart.stories.tsx
```diff
@@ -305,7 +303,7 @@ const UpdatingChartValues = () => {
           domain: { max: 250 },
         }}
       >
-        <ReferenceLine LineComponent={MyCustomLine} dataY={0} />
+        <ReferenceLine LineComponent={SolidLine} dataY={0} />
```
hcopp, line 306: Realized our goal here to allow animations on the line wasn't realistic, we use a DerivedValue in CDS so it doesn't animate value changes. We use DerivedValue for ReferenceLine so it can be used by Scrubber. https://github.com/coinbase/cds/blob/master/packages/mobile-visualization/src/chart/Path.tsx#L172

This logic would have worked on web though, which is why we likely didn't think about it in the past when porting.

packages/mobile/src/visualizations/chart/Path.tsx
```diff
@@ -120,6 +120,7 @@ export type PathProps = PathBaseProps &
     transition?: Transition;
     /**
      * The SVG path data string.
+     * @note d only supports transitions for string values, not animated values.
```
hcopp, line 123: Clarifying the discovery from above

packages/mobile/src/visualizations/chart/line/__stories__/LineChart.stories.tsx
```diff
@@ -1995,6 +1997,10 @@ function ExampleNavigator() {
         title: 'Multiple Lines',
         component: <MultipleLine />,
       },
+      {
+        title: 'Multiple Lines No Animations',
+        component: <MultipleLine animate={false} />,
+      },
```
hcopp, line 2003: This lets us test out instant enter charts, key to match some of our customers.

packages/mobile/src/visualizations/chart/CartesianChart.tsx
```diff
@@ -69,6 +73,33 @@ const ChartCanvas = memo(
...
+  }, []);
+
+  useEffect(() => {
+    return () => {
+      if (containerLayoutRafRef.current !== null) cancelAnimationFrame(containerLayoutRafRef.current);
+    };
+  }, []);
+
+  return [containerLayout, onContainerLayout];
+}
```
hcopp, line 98: Now onLayout is synchronous, but this breaks our chart logic across the board. The simplest thing I have found so far is to simply chart size in the same manner we did in v8.

packages/mobile/src/visualizations/chart/scrubber/ScrubberBeaconLabelGroup.tsx
```diff
@@ -271,35 +287,53 @@ export const ScrubberBeaconLabelGroup = memo<ScrubberBeaconLabelGroupProps>(
...
-      const dimensions = validPositions.map((pos) => {
-        const trackedDimensions = labelDimensions[pos.seriesId];
-        return {
-          seriesId: pos.seriesId,
-          width: trackedDimensions?.width ?? maxLabelWidth,
-          height: trackedDimensions?.height ?? maxLabelHeight,
-          preferredX: pos.x,
-          preferredY: pos.desiredY,
-        };
-      });
```
hcopp, line 290: Before we got this glitch on a chart with animations disabled

https://github.com/user-attachments/assets/0e887b35-d085-42d2-a58a-0f745ad5de71

Now we don't render (keep returning null) until we have properly tracked the location, and thus know how much space to leave between each label and which side it can be rendered on.

hcopp: This wasn't noticeable when animations were enabled

packages/mobile/src/visualizations/chart/scrubber/ScrubberBeaconLabelGroup.tsx
```diff
@@ -56,21 +56,37 @@ const PositionedLabel = memo<{
...
       (current, previous) => {
+        if (current.position === null || current.position === undefined) {
+          idleAnimatedY.value = null;
+          return;
+        }
+        // Do not spring from placeholder targetY (0 while no position) when the label first lays out.
+        if (previous?.position === null || previous?.position === undefined) {
+          idleAnimatedY.value = current.y!;
+          return;
+        }
```
hcopp, line 84: When we first calculate the values we shouldn't animate still

adrienzheng-cb (approved):

PR #786, opened 2026-07-10
feat: add createIcon factory to decouple Icon from a single icon set

adrienzheng-cb:
> **Example PR** demonstrating the `createIcon` seam ("Option B") from the [CDS Icon Style Update TDD](https://linear.app/coinbase/document/tdd-cds-icon-style-update-for-retail-app-refresh-90c56f987b9c). Backward-compatible infrastructure only — no new icon assets.

## What changed? Why?

The web and mobile `Icon` components hardcode a single icon source (`@coinbase/cds-icons/glyphMap`), a single font family (`'CoinbaseIcons'`), and a single `name` type (`IconName`). That makes it impossible to render an alternate icon set (e.g. a private, retail-specific style) without forking the whole component.

This PR extracts the rendering logic into a **`createIcon<Name>` factory** (mirroring the existing `createIllustration<Variant>` convention) that binds:

- a **glyph map** (`GlyphMap<Name>`, keys `` `${name}-${size}-${active}` ``),
- an injectable **font family** (defaults to `'CoinbaseIcons'`),
- a typed **name union** (`Name`), and
- an optional **glyph resolver** (`getGlyph`) for sets with a different key format or size model.

The default `Icon` becomes a one-liner:

```ts
export const Icon = createIcon<IconName>({ glyphMap, fontFamily: DEFAULT_ICON_FONT_FAMILY });
```

So the public API and behavior are **unchanged**, while a consumer with their own icon package can do `createIcon<RetailIconName>({ glyphMap, fontFamily })` and get a fully-typed icon component that reuses all CDS accessibility, theming, sizing, and `componentConfig` behavior — no duplicated component logic.

### Glyph resolution

Rather than a narrow `getSourceSize` hook that only remapped the pixel size, the factory accepts a general **`getGlyph`** resolver: `(args: IconGlyphResolverArgs<Name>) => string | undefined`, where args carry `glyphMap`, `name`, `size` token, resolved `pixelSize`, and `active`. The default resolver keeps the CDS `` `${name}-${sourceSize}-${'active' | 'inactive'}` `` scheme (`sourceSize` = 12/16/24), so nothing changes for the default `Icon`; consumers whose maps are keyed differently can override the entire lookup.

### Font family (web)

On web, `font-family` is applied through a `--cds-icon-font-family` CSS custom property (default `'CoinbaseIcons'`, declared in the static Linaria block). `createIcon` only sets the variable inline when a non-default font is bound, so the default keeps its fully static CSS while consumers can override the font per instance via `className`/`styles` or by scoping the variable on an ancestor. Mobile applies `fontFamily` via its inline style object (React Native has no CSS variables).

### Docs

The Icon docs page gains a **"Custom icon sets"** section (web + mobile) documenting the factory, the glyph map contract, the `getGlyph` override, and font loading. The web page includes an **editable live demo** that binds `createIcon` to Google Material Icons to prove a second, fully-typed icon set works end-to-end.

### Storybook

Adds a **`CreateIcon` web story** (`packages/web/src/icons/__stories__/CreateIcon.stories.tsx`) that renders the `createIcon` + Google Material Icons demo — a second, fully-typed icon component reusing the CDS Icon renderer (sizing, color, accessibility) with a different font and glyphs. The story is excluded from web visual regression (Percy) in `.percy.js` because it loads an externally-hosted font that can change over time and break snapshots.

No equivalent mobile story is included: React Native fonts must be bundled via `expo-font` (see `apps/expo-app/src/hooks/useFonts.ts`) and require a native rebuild, so the web pattern of loading a network font in a story doesn't translate. The mobile factory is covered by unit tests and the mobile docs section instead; a visual mobile demo can be a follow-up if we bundle a second font.

N/A — additive infrastructure.

## UI changes

None. The default icon set, font, glyph keys, and sizing are identical, so rendering is unchanged.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

### Testing instructions

- `yarn nx run-many --target=typecheck --projects=web,mobile` — pass
- `yarn nx run web:test --testPathPattern="createIcon.test"` — pass (renders glyph, sets/omits the font-family CSS variable, renders fallback)
- `yarn nx run mobile:test --testPathPattern="createIcon.test"` — pass (renders active/inactive glyph, applies/defaults font family, renders fallback, custom `getGlyph` resolver)
- `yarn nx run docs:start` — docs build succeeds and the Icon page renders the live demo

N/A — no files under `packages/icons/**` or `packages/illustrations/**` changed; no icon assets or names added.

## Change management

type=routine
risk=low
impact=sev5

automerge=false

### DIFF
--- a/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
+++ b/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
@@ -29,15 +29,14 @@ Mobile apps cache font files in the native build, so you must rebuild your app a
 
 ### Custom icon sets
 
-CDS ships a single default icon font (`@coinbase/cds-icons`). If you maintain your own icon package — with its own generated glyph map, font, and name union — use `createIcon` to build a fully-typed icon component bound to that set. It reuses all of the default `Icon`'s rendering, sizing, color, and accessibility behavior; only the glyphs, font, and `name` type change. In fact, the default `Icon` is itself `createIcon<IconName>({ glyphMap, fontFamily: 'CoinbaseIcons' })`, so this API is fully backward compatible.
+Most apps only need the default `Icon`. If you have your own icon font, `createIcon` builds a matching, fully-typed `Icon` for it that reuses the same sizing, color, and accessibility behavior — only the glyphs, font, and `name` type change. Support for building your own icon set is expected to become more self-service over time.
 
 ```jsx
 import { createIcon } from '@coinbase/cds-mobile/icons';
-// Your icon package's generated glyph map + name union
+// Your icon font's glyph map and icon-name type
 import { glyphMap } from '@my-org/icons/glyphMap';
 import type { MyIconName } from '@my-org/icons';
 
-// `name` is typed to MyIconName and behaves exactly like the default Icon
 export const MyIcon = createIcon<MyIconName>({
   glyphMap,
   fontFamily: 'MyIcons',
--- a/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
+++ b/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
@@ -49,3 +49,5 @@ export const MyIcon = createIcon<MyIconName>({
 ```
 
 The `glyphMap` must be keyed by `` `${name}-${size}-${active | inactive}` `` (where `size` is `12`, `16`, or `24`), with each value the single font glyph character — the same shape as `@coinbase/cds-icons`. Load your font before rendering (for example with `expo-font`'s `useFonts`) using the same family name you pass as `fontFamily`. Because mobile bundles fonts into the native build, rebuild your app after adding a new font.
+
+If your icon set uses a different key format or size model, pass a `getGlyph` resolver to `createIcon` to control exactly how a glyph is looked up from the map.
--- a/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
+++ b/apps/docs/docs/components/media/Icon/_mobileExamples.mdx
@@ -26,3 +26,26 @@ Mobile apps cache font files in the native build, so you must rebuild your app a
   </HStack>
 </VStack>
 ```
+
+### Custom icon sets
+
+CDS ships a single default icon font (`@coinbase/cds-icons`). If you maintain your own icon package — with its own generated glyph map, font, and name union — use `createIcon` to build a fully-typed icon component bound to that set. It reuses all of the default `Icon`'s rendering, sizing, color, and accessibility behavior; only the glyphs, font, and `name` type change. In fact, the default `Icon` is itself `createIcon<IconName>({ glyphMap, fontFamily: 'CoinbaseIcons' })`, so this API is fully backward compatible.
+
+```jsx
+import { createIcon } from '@coinbase/cds-mobile/icons';
+// Your icon package's generated glyph map + name union
+import { glyphMap } from '@my-org/icons/glyphMap';
+import type { MyIconName } from '@my-org/icons';
+
+// `name` is typed to MyIconName and behaves exactly like the default Icon
+export const MyIcon = createIcon<MyIconName>({
+  glyphMap,
+  fontFamily: 'MyIcons',
+});
+```
+
+```jsx
+<MyIcon name="rocket" size="l" active />
+```
+
+The `glyphMap` must be keyed by `` `${name}-${size}-${active | inactive}` `` (where `size` is `12`, `16`, or `24`), with each value the single font glyph character — the same shape as `@coinbase/cds-icons`. Load your font before rendering (for example with `expo-font`'s `useFonts`) using the same family name you pass as `fontFamily`. Because mobile bundles fonts into the native build, rebuild your app after adding a new font.
--- a/packages/mobile/src/icons/createIcon.tsx
+++ b/packages/mobile/src/icons/createIcon.tsx
@@ -31,6 +31,20 @@ export type GlyphMap<Name extends string> = Record<
   string
 >;
 
+/** Arguments passed to a glyph resolver to look up an icon's glyph. */
+export type IconGlyphResolverArgs<Name extends string> = {
+  /** Glyph map bound to this icon set. */
+  glyphMap: GlyphMap<Name>;
+  /** Icon name requested via the `name` prop. */
+  name: Name;
+  /** Size token requested via the `size` prop. */
+  size: IconSize;
+  /** Resolved pixel size from the theme (includes device font scaling). */
+  pixelSize: number;
+  /** Whether the active variant was requested. */
+  active: boolean;
+};
+
 /** Configuration used to bind an icon set to a typed `Icon` component. */
 export type CreateIconConfig<Name extends string> = {
   /** Generated glyph map for this icon set. */
@@ -41,10 +55,12 @@ export type CreateIconConfig<Name extends string> = {
    */
   fontFamily?: string;
   /**
-   * Maps a resolved pixel size to the source glyph size. Override for icon
-   * sets that use a different size model.
+   * Resolves the glyph to render for an icon from the glyph map. Override to
+   * support a custom key format or size model. Defaults to the CDS scheme:
+   * `${name}-${sourceSize}-${'active' | 'inactive'}`, where `sourceSize` is
+   * `12`, `16`, or `24`.
    */
-  getSourceSize?: (iconSize: number) => IconSourcePixelSize;
+  getGlyph?: (args: IconGlyphResolverArgs<Name>) => string | undefined;
 };
 
 export type IconBaseProps<Name extends string = string> = SharedProps &
@@ -96,6 +112,18 @@ const getIconSourceSize = (iconSize: number): IconSourcePixelSize => {
   return 24;
 };
 
+/** Default glyph resolver using the CDS `${name}-${sourceSize}-${state}` key scheme. */
+const defaultGetGlyph = <Name extends string>({
+  glyphMap,
+  name,
+  pixelSize,
+  active,
+}: IconGlyphResolverArgs<Name>): string | undefined => {
+  const sourceSize = getIconSourceSize(pixelSize);
+  const key = `${name}-${sourceSize}-${active ? 'active' : 'inactive'}` as keyof GlyphMap<Name>;
+  return glyphMap[key];
+};
+
 /**
  * Creates a typed `Icon` component bound to a specific icon set (glyph map,
  * font family, and name union). The default CDS `Icon` is created from this
@@ -106,7 +134,7 @@ const getIconSourceSize = (iconSize: number): IconSourcePixelSize => {
 export function createIcon<Name extends string>({
   glyphMap,
   fontFamily = DEFAULT_ICON_FONT_FAMILY,
-  getSourceSize = getIconSourceSize,
+  getGlyph = defaultGetGlyph,
 }: CreateIconConfig<Name>) {
   const Icon = memo((_props: IconProps<Name>) => {
     const mergedProps = useComponentConfig('Icon', _props);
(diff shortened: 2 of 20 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

apps/docs/docs/components/media/Icon/_mobileExamples.mdx
```diff
@@ -26,3 +26,26 @@ Mobile apps cache font files in the native build, so you must rebuild your app a
   </HStack>
 </VStack>
 ```
+
+### Custom icon sets
+
+CDS ships a single default icon font (`@coinbase/cds-icons`). If you maintain your own icon package — with its own generated glyph map, font, and name union — use `createIcon` to build a fully-typed icon component bound to that set. It reuses all of the default `Icon`'s rendering, sizing, color, and accessibility behavior; only the glyphs, font, and `name` type change. In fact, the default `Icon` is itself `createIcon<IconName>({ glyphMap, fontFamily: 'CoinbaseIcons' })`, so this API is fully backward compatible.
```
cb-ekuersch: > with its own generated glyph map, font, and name union

This comes across as kind of confusing without knowing what any of them are. 

I wonder if it would just be better to not document createIcon until we have a better way to document all the required arguments and make them a little more self-service? Otherwise i think we need to simplify by taking out some technical details and simply showing how createIcon is used. We could say in the future it may be possible to build your own icon set

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,217 @@
...
+};
+
+/**
+ * Creates a typed `Icon` component bound to a specific icon set (glyph map,
+ * font family, and name union). The default CDS `Icon` is created from this
+ * factory; consumers with their own icon package can create their own typed
+ * icon component that reuses all of the CDS rendering, accessibility, and
+ * theming behavior.
+ */
+export function createIcon<Name extends string>({
```
cb-ekuersch, line 135: need unit tests like web

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,217 @@
...
+    const iconSize = theme.iconSize[size] * fontScale;
+    const sourceSize = getSourceSize(iconSize);
+
+    const iconColor = theme.color[color];
+    const finalColor = dangerouslySetColor ?? iconColor;
+
+    const rootStyle = useMemo(
+      () => [
+        {
+          paddingTop: theme.space[paddingTop ?? paddingY ?? padding ?? 0],
```
cb-ekuersch, line 176: this is really interesting... i wonder what other `_props` should actually be going to the Box contianer and not the inner Text. I see we are not spreading anywhere so hopefully the set is small

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,217 @@
...
+    const TextComponent = animated ? Animated.Text : Text;
+    const theme = useTheme();
+    const { fontScale } = useWindowDimensions();
+
+    // Scale according to device a11y font size settings
+    const iconSize = theme.iconSize[size] * fontScale;
+    const sourceSize = getSourceSize(iconSize);
+
+    const iconColor = theme.color[color];
+    const finalColor = dangerouslySetColor ?? iconColor;
```
cb-ekuersch, line 171: nit: make sure this prop has a deprecation date

cb-ekuersch: if so, we could probably consider dropping it since the color prop covers this and styles.icon.color should work as well

cb-ekuersch: dropping as in the business logic, keeping the deprecated prop for bw compat but have it wired up to nothgin

packages/web/src/icons/__stories__/CreateIcon.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import { HStack, VStack } from '../../layout';
+import { Text } from '../../typography/Text';
+import { createIcon, type GlyphMap } from '../createIcon';
+import { Icon } from '../Icon';
+
+/**
+ * `createIcon` bound to a different font (Google Material Icons), showing a
+ * consumer can supply their own glyph map, font, and name union while reusing
+ * the CDS Icon renderer. Uses a network font, so percy is skipped.
```
cb-ekuersch: what do you mean percy is skipped?

apps/docs/docs/components/media/Icon/_mobileExamples.mdx
cb-ekuersch: maybe there is a helpful doc about glyphmap icons to link to for supplemental reading

packages/web/src/icons/__stories__/CreateIcon.stories.tsx
adrienzheng-cb: we don't include this story in visreg, because the font comes from a external source and may change and break the test.

adrienzheng-cb: I'm gonna remove this comment tho, it is too trivial a detail.

packages/mobile/src/icons/createIcon.tsx
adrienzheng-cb: Good call. It already has `@deprecationExpectedRemoval v10`. I'd prefer to keep the current wiring for now rather than turn it into a no-op early, since that's a behavior change for existing consumers still passing `dangerouslySetColor`. Plan is to drop the business logic (and/or the prop) as part of the v10 removal, where `color` and `styles.icon.color` fully cover it.

packages/mobile/src/icons/createIcon.tsx
adrienzheng-cb: This is intentional — mobile `createIcon` destructures a fixed, closed set of props and never spreads `...props` onto the `Box` or the inner `Text`, so nothing leaks through. Anything not in that set is simply not accepted. If we later want web-style `Box` passthrough on mobile, that'd be a separate, deliberate change.

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,249 @@
...
+  type TextStyle,
+  useWindowDimensions,
+  type ViewStyle,
+} from 'react-native';
+import type { ThemeVars } from '@coinbase/cds-common/core/theme';
+import type { IconSize, IconSourcePixelSize } from '@coinbase/cds-common/types/IconSize';
+import type { SharedAccessibilityProps } from '@coinbase/cds-common/types/SharedAccessibilityProps';
+import type { SharedProps } from '@coinbase/cds-common/types/SharedProps';
+import type { PaddingProps } from '@coinbase/cds-common/types/SpacingProps';
+import { isDevelopment } from '@coinbase/cds-utils';
```
cb-ekuersch, line 15: it would be great if we could stop relying on this package. Also how does it perform a development check? Using NODE_ENV could be brittle as not every build tool is node-centric these days

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,249 @@
...
+     * @default primary
+     */
+    color?: ThemeVars.Color;
+    /**
+     * @deprecated Use `style`, `styles.icon`, or the `color` prop to customize icon color. This will be removed in a future major release.
+     * @deprecationExpectedRemoval v10
+     */
+    dangerouslySetColor?: string | Animated.AnimatedInterpolation<string>;
+    animated?: boolean;
+    style?: Animated.WithAnimatedValue<StyleProp<TextStyle>>;
```
cb-ekuersch: I dont think we typically put style on BaseProps anymore, right?

packages/mobile/src/icons/createIcon.tsx
```diff
@@ -0,0 +1,249 @@
...
+ * factory; consumers with their own icon package can create their own typed
+ * icon component that reuses all of the CDS rendering, accessibility, and
+ * theming behavior.
+ */
+export function createIcon<Name extends string>({
+  glyphMap,
+  fontFamily = DEFAULT_ICON_FONT_FAMILY,
+  getGlyph = defaultGetGlyph,
+}: CreateIconConfig<Name>) {
+  const Icon = memo((_props: IconProps<Name>) => {
```
cb-ekuersch: I noticed we forward a ref on web, but for react-native we dont (we can use ref as a prop instead of forwardRef as well)

adrienzheng-cb: Good parity catch. Added `ref` forwarding on mobile using the ref-as-prop pattern (matching our `Text` component, which does the same rather than `forwardRef`) — it forwards to the inner glyph `Text`, mirroring web forwarding to the glyph `span`. Also added a ref-forwarding unit test. Done in 4731de0a8.

packages/mobile/src/icons/createIcon.tsx
adrienzheng-cb: Good catch — moved `style` off `IconBaseProps` onto `IconProps`, which also matches the web factory (its `IconBaseProps` has no `style`; it comes in via the full props). Done in 4731de0a8.

packages/mobile/src/icons/createIcon.tsx
adrienzheng-cb: `isDevelopment()` is the shared CDS convention — it's used by the sibling `createIllustration.tsx` as well as `Control.tsx`, `TextIcon.tsx`, `RadioGroup`/`CheckboxGroup`, etc. Under the hood (`packages/utils/src/env.ts`) it does `process.env.NODE_ENV === 'development'`, so you're right that it assumes a node-centric build. I'd prefer not to diverge just this one file here; for the mobile package specifically `__DEV__` (the Metro global) would be the more correct/robust check. Happy to open a follow-up to standardize the dev-check across the mobile package rather than change it piecemeal in this PR — keeping it consistent with `createIllustration` for now.

adrienzheng-cb: separate ticket to track https://linear.app/coinbase/issue/CDS-2513/standardize-development-environment-check-in-packagesmobile-dev-vs

packages/mobile/src/icons/createIcon.tsx
adrienzheng-cb: added.

cb-ekuersch (approved):

PR #848, opened 2026-08-18
feat: support custom icon sets in name-prop components

adrienzheng-cb:
> **Phase 1 of 2.** This PR ships only the *runtime* half of custom icon set support: the glyph-source provider. The compile-time name registry that earlier revisions of this branch contained has been removed and deferred to phase 2 — see [Not in this PR](#not-in-this-pr).
>
> Design doc: **[Custom icon sets in CDS name-prop components](https://linear.app/coinbase/document/custom-icon-sets-in-cds-name-prop-components-proposal-955525899ee5)**

### The problem

A large number of CDS components render an icon from a **name** rather than from a node: `IconButton`, `Button`, `Tag`, `Cell`, `Banner`, `ModalHeader`, `Toast`, `Calendar`, and many more. The name is looked up in the CDS glyph map and drawn in the CDS icon font.

A team that ships its own icon font — concretely `@cbhq/retail-icons`, which is a restyle of the CDS set — had no way to make those components draw *its* glyphs. The only options were forking each component or wrapping every call site in a bespoke factory.

### The change: inject a glyph *source*, not components

`IconGlyphSourceProvider` adds a **glyph source** to every CDS icon rendered beneath it:

```tsx
<IconGlyphSourceProvider source={retailIconSource}>
  {/* unmodified CDS IconButton, now drawing the retail "close" in the retail font */}
  <IconButton name="close" onPress={dismiss} />
</IconGlyphSourceProvider>
```

A source is exactly the object `createIcon` already accepts — `{ glyphMap, fontFamily?, getGlyph? }` — so an icon package exposes **one** object that serves both roles: build a standalone typed `Icon` from it, or inject it as a source. That is now literally one type: `CreateIconConfig` and `IconGlyphSource` described the same shape in two modules, so they are collapsed into a single exported `IconGlyphSource<Name extends string = string>`, and `createIcon`'s parameter is named `source` to match.

**How resolution works.** Glyph lookup in `createIcon` moved into a small `resolveGlyph` helper. It tries the source from context first, then the set the component was bound to, and returns the first match together with *that source's* font:

```ts
const resolved = resolveGlyph(contextSource, source, { name, size, pixelSize, active });
```

Each source is asked using **its own** `getGlyph` (falling back to the CDS key scheme) against **its own** `glyphMap`. If a source has no glyph for the request it returns `undefined` and resolution falls through to the next one, with the component's bound set as the final fallback. If nothing resolves, behavior is exactly as before: a dev-only `console.error` and the `fallback` prop.

**Exactly one custom source is ever in effect.** The context holds a single `IconGlyphSource | undefined`, not a list, and a nested provider *replaces* the outer source rather than layering on top of it — so the hook is `useIconGlyphSource`, singular. This follows the same decision as the rest of the PR: composing several icon sets at once is not a case we support, so the context does not pay for it. Note this narrows but does not remove the mixed-font caveat — a source can cover a name at one size or state and not another, so a single icon can still resolve partly from the custom source and partly from the built-in set, in two different fonts.

Two consequences worth calling out:

- **A source brings its own key scheme and its own font.** Because `getGlyph` and `fontFamily` travel with the source rather than with the component, a set keyed as `` `${name}-${'s'|'l'}-${state}` `` in a completely different font composes inside the same `IconButton` as the CDS set, which is keyed `` `${name}-${12|16|24}-${state}` ``. The custom source and the built-in set resolve per-icon, each in its own font. On web the matching font is applied through the `--cds-icon-font-family` CSS variable (only when it differs from the default); on mobile it goes straight into the `Text` style.
- **It reaches icons CDS renders internally.** Roughly 60 call sites across web and mobile hardcode names like `close`, `caretUp`, and `checkmark` inside components — the modal close button, `CellAccessory`, `Calendar`, `SegmentedControl`, `TableCellSortIcon`, and so on. Those all render the same `Icon`, and `Icon` consults the context itself, so they pick up the custom set with **zero changes to any of them**. No prop-based approach (`icon={<CustomIcon />}`, `IconComponent={…}`) can do this: it would require threading a prop through every one of those call sites and every component that wraps them.

### Prerequisite for consumers: scope the provider to a subtree, not the app root

> [!WARNING]
> **This is the one thing adopters must get right.** The provider re-skins *every* CDS icon below it whose name the custom source also provides — including all of the internal call sites described above. Registering it at the app root therefore restyles far more than intended.
>
> Measured in the retail app: **62 non-navigation call sites** would be affected by a root-level registration, **41 of them for `coinbaseOne` alone**. Wrap only the subtree that actually wants the restyle (a tab, a flow, a screen).

This is a usage caveat, not a code change in this PR; it is documented on the provider's JSDoc and called out in the docs entry described below.

### `createIcon` stays; the provider joins it

Both mechanisms ship, because they answer different questions. `createIcon` builds a **separate, typed component** for a set that is its own thing; `IconGlyphSourceProvider` changes the glyphs that the components you **already render** draw. Removing the factory was considered during review and reversed — it stays exactly as it is, still public and still exported from both `icons` barrels, and the Material Icons docs example and `CreateIcon.stories.tsx` are untouched.

What was missing is that the provider had **no docs entry and no story at all** — the Icon docs only ever covered `createIcon` — so the mechanism this PR is actually about was undiscoverable. This PR adds, alongside the existing `createIcon` material rather than in place of it:

- a provider section in the web and mobile Icon docs, framed against `createIcon` so a reader can tell which of the two they want, including the subtree-scoping warning above;
- a live web demo where unmodified `IconButton`s render Material Icons glyphs inside the provider and CDS glyphs immediately outside it;
- `IconGlyphSourceProvider.stories.tsx`, the same comparison in Storybook.

The Material Icons fixture is now shared between the two stories instead of duplicated, and the new story is Percy-excluded for the same network-font non-determinism reason as the existing one.

### Why no type changes are needed in phase 1

A custom icon set is a **restyle of icons CDS already names**. `close`, `home`, and `coinbaseOne` are all in `IconName` today, so `<IconButton name="close" />` type-checks inside the provider with no widening whatsoever. Phase 1 therefore touches **no** public type:

- `IconName` is unchanged.
- `Icon` and `IconButton` still accept `IconName` on both platforms.
- `createIcon`'s signature is unchanged — still `createIcon<Name extends string>`.

That last one is deliberate. Constraining it to `createIcon<Name extends IconName>` was considered and **rejected**: it would break CDS's documented "bring your own icon font" contract, including the live Material Icons example in the docs (`apps/docs/docs/components/media/Icon/_webExamples.mdx` plus its mobile twin) and `CreateIcon.stories.tsx`, all of which bind names like `favorite` and `delete` that are not CDS icon names. The invariant that a custom set uses CDS names will be enforced where the set is generated (in the `retail-icons` sync script), not in CDS's types.

### Phase 2 (deferred)

If retail ever needs an icon with **no** CDS counterpart, phase 2 adds a compile-time registry back: a global `Cds.IconNameRegistry` interface that an icon package augments, and a `RegisteredIconName = IconName | Cds.IconNameRegistry[keyof …]` that component props accept.

That change is purely **additive** — it widens an existing union — so deferring it costs nothing and **nothing shipped in phase 1 has to be migrated** to get there. Doing it now would mean shipping and reviewing a global type-system escape hatch for a need that does not yet exist.

### Not in this PR

Deliberately excluded, not overlooked:

| Excluded | Why |
| --- | --- |
| The `Cds.IconNameRegistry` type registry / `RegisteredIconName` | Not needed while custom sets restyle existing CDS names. Deferred to phase 2, additively. |
| A `createIcon<Name extends IconName>` constraint | Rejected — breaks the documented bring-your-own-icon-font contract, the Material Icons docs example, and `CreateIcon.stories.tsx`. |
| Name-mapping work (reconciling a custom set's names against CDS names) | Belongs in the `retail-icons` sync script, not in CDS. |

N/A — feature.

## UI changes

**No visual change to any existing component.** With no provider mounted, the resolution order, rendering, accessibility, and theming behavior are exactly what they were; the new capability is purely additive. Screenshot tables are omitted for that reason. The only new UI is the provider story and docs demo, both of which exist to show the difference between an icon inside the provider and the same icon outside it.

For reference, an unmodified `IconButton` rendering `@cbhq/retail-icons` glyphs alongside built-in CDS glyphs was verified locally in `retail-icons-storybook`. That Storybook wiring points at a sibling checkout, so it is not part of this PR.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [x] Manual - Web

Mobile is covered by unit tests only. It is a line-for-line port of the web change, and RN actually takes the simpler path — the resolved source's `fontFamily` goes directly into the `Text` style, with no CSS-variable indirection. Worth a device check before this ships if a reviewer wants one.

### Testing instructions

New `createIcon` tests on both platforms cover the two behaviors that matter: resolving a name that **only** a provider source owns (asserting the source's font is applied), and a provider source **overriding** a name the bound set already owns. Existing tests were untouched — unifying the two config types renamed a type, not a behavior.

```
yarn nx run web:test --testPathPattern=icons/__tests__/createIcon
yarn nx run mobile:test --testPathPattern=icons/__tests__/createIcon
yarn nx run-many --target=typecheck --projects=web,mobile,common
yarn nx run-many --target=typecheck --projects=storybook,expo-app,docs
```

To exercise it manually, wrap a subtree in `IconGlyphSourceProvider` with a second icon font and render `<IconButton name="…" />` using a name that font provides.

N/A — no files under `packages/illustrations/**` or `packages/icons/**` change. The icon font and glyph map are untouched.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

apps/docs/docs/components/media/Icon/_mobileExamples.mdx
```diff
@@ -29,7 +29,16 @@ Mobile apps cache font files in the native build, so you must rebuild your app a
 
 ### Custom icon sets
 
-Most apps only need the default `Icon`. If you have your own icon font, `createIcon` builds a matching, fully-typed `Icon` for it that reuses the same sizing, color, and accessibility behavior — only the glyphs, font, and `name` type change. Support for building your own icon set is expected to become more self-service over time.
+Most apps only need the default `Icon`. There are two ways to bring your own glyphs, and they solve different problems:
+
+- **`createIcon`** builds a separate, fully-typed icon component for your font. Use it when your icons are their own set, rendered through your own component.
+- **`IconGlyphSourceProvider`** adds your glyphs to the icons CDS already renders. Use it when you want the components you're already using — `IconButton`, `Button`, `Tag` — to draw their icons from your font.
```
cb-ekuersch: probalby should note here that there is an expectation the icon names match CDS' icons

packages/mobile/src/icons/IconGlyphSourceContext.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { createContext, useContext } from 'react';
+
+import type { IconGlyphSource } from './createIcon';
+
+const IconGlyphSourceContext = createContext<IconGlyphSource<any> | undefined>(undefined);
+
+/** The nearest ancestor provider's glyph source, if any. */
+export function useIconGlyphSource(): IconGlyphSource<any> | undefined {
+  return useContext(IconGlyphSourceContext);
```
cb-ekuersch: should this return a default if there is no provider? otherwise this will return undefined as the signature points out

packages/mobile/src/icons/IconGlyphSourceContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export type IconGlyphSourceProviderProps = {
```
cb-ekuersch: nit: do you think we need to export this? Just thinking of all the things our customers have picked up without our consent

packages/mobile/src/icons/IconGlyphSourceContext.tsx
adrienzheng-cb: The `undefined` is intentional and load-bearing. Every `Icon` is created through `createIcon` bound to a source at creation time (the default CDS `Icon` is bound to the CDS glyph map), so the component always has a fallback set of its own. The provider is only an optional override layered on top, and `undefined` from this hook is what tells `resolveGlyph` "no override here, use the bound source."

That also means there's no single default we could return instead: the correct fallback differs per component (CDS's `Icon` should fall back to CDS, a consumer's `createIcon(...)` should fall back to their own set), and only the component knows its own bound source — the hook doesn't. Returning a fabricated default here would flatten the two-tier resolution into one.

So the flow is: provider source for that glyph → else the component's bound source → else dev-only `console.error` + render the `fallback` prop. I've reworded the hook's doc line to say this explicitly.

packages/mobile/src/icons/IconGlyphSourceContext.tsx
adrienzheng-cb: the consumers will need to use the `IconGlyphSourceProvider` in their code to get this feature so I assume exporting the prop type is necessary as well.

however, I removed export for the useIconGlyphSource hook since it only needs to be used by our createIcon factory/Icon component.

cb-ekuersch (approved):

PR #849, opened 2026-08-18
POC: generic `IconComponent` prop for custom icon sets (Alternative 1, do not merge)

adrienzheng-cb:
> [!IMPORTANT]
> **This is a proof of concept — do not merge.** It exists to inform the design discussion in [Custom icon sets in CDS name-prop components — proposal](https://linear.app/coinbase/document/custom-icon-sets-in-cds-name-prop-components-proposal-955525899ee5). It implements **Alternative 1 (generic `IconComponent` prop)**, which the doc records as *not* adopted, so we can compare it against the adopted `IconGlyphSourceProvider` + `Cds.IconNameRegistry` approach on equal footing. Branched from `master` so it stands alone and does not depend on the provider work.

## What changed? Why?

CDS components that take an icon by name string type that prop as `IconName`, a closed union of built-in CDS icon names. A team shipping its own icon set (`@cbhq/retail-icons`, built with `createIcon()`) cannot pass its own names: the name isn't in the union, and the component internally renders the built-in `Icon`, which has no glyph for it.

This POC adds an optional **`IconComponent`** prop. The component's props become generic over it, and the `name` prop's accepted values are **derived from whatever icon component was passed**, defaulting to the built-in `Icon` when the prop is omitted.

```tsx
// 1. unchanged — still type-checks against the built-in IconName
<IconButton name="close" />

// 2. name is now checked against RetailIcon's own name union
<IconButton IconComponent={RetailIcon} name="navSearch" />

// 3. type ERROR — not a retail glyph name
<IconButton IconComponent={RetailIcon} name="close" />
```

The value over the adopted approach is that type safety is **per-call-site rather than global**: no global type widening, and no possibility of a name being accepted in a subtree where its glyph doesn't exist. Case 3 is the interesting one — a global name registry cannot express it, because once `navSearch` is registered it is valid everywhere, and `close` stays valid even in a subtree whose provider only supplies retail glyphs.

### Scope

| Component | Platforms | Why it's here |
| --- | --- | --- |
| `IconButton` | web + mobile | The canonical example from the proposal. Exercises the hard parts: icon size derived from its own `size` prop, icon color resolved from its variant, `ProgressCircle` in the loading state, flush margins. |
| `Banner` | web + mobile | The second component, chosen to stress the pattern differently (see below). |

I deliberately did **not** attempt the rest of the library. See [Rollout cost](#what-would-it-cost-to-roll-out) for what the remaining components would take.

**Why `Banner` as the second component** — it stresses three things `IconButton` does not:

1. **`startIcon` is required**, not optional, so the derived-name machinery has to work for a required prop with no fallback.
2. **It is not polymorphic.** `IconButton` already had a generic call signature for `as`, so the icon set could ride along as a second type parameter. `Banner` had no generic at all, so one has to be introduced *from scratch* purely to carry the icon set. That is the situation for the large majority of CDS components, which makes `Banner` the more representative sample of rollout cost.
3. **It renders a second icon the prop cannot reach** — the dismiss button's `close` glyph. That turns the approach's central limitation into something concrete and testable rather than theoretical, and there is a passing test asserting exactly this mixed-font outcome.

`DefaultCarouselNavigation` (four independent name props) and `Stepper` (five state-keyed names) were the other candidates. I note below what implementing them would involve; I judged `Banner` more informative because the multi-name question has a fairly predictable answer, whereas "what does this cost on a non-polymorphic component, and what does it fail to reach" does not.

## How the `forwardRef` / `memo` generic-erasure problem was solved

Both wrappers take a *concrete* props type, so wrapping a generic render function collapses the generic to a single instantiation and the derivation silently stops working. The fix is the standard one — declare the public type as a **generic call signature** and cast the wrapped component to it:

```ts
type IconButtonComponent = (<
  AsComponent extends React.ElementType = IconButtonDefaultElement,
  IconComponentType extends IconLike = typeof Icon,
>(
  props: IconButtonProps<AsComponent, IconComponentType>,
) => Polymorphic.ReactReturn) &
  Polymorphic.ReactNamed;

export const IconButton = memo(
  forwardRef<React.ReactElement<IconButtonBaseProps>, IconButtonBaseProps>(/* ... */),
) as unknown as IconButtonComponent;
```

Two findings worth recording, both of which cost me a round trip:

- **Making the render function itself generic does not work.** My first attempt gave the inner function both type parameters. `forwardRef` instantiates type parameters with their constraints during inference, so `IconComponentType` collapsed to `IconLike`, and the polymorphic `Omit`/`ExtendableProps` machinery then produced a props type that had lost `name` entirely:

  > `Property 'name' is missing in type 'Omit<IconButtonProps<ElementType, IconLike>, "ref">' but required in type '{ name: IconName; ... }'`

  The working shape leaves the **implementation non-generic** — it sees the default instantiation and only needs to forward `name` to whichever component it was handed. The generic exists *only* in the exported call signature. This is fine but worth being explicit about: the generic is a pure API-surface construct with no representation in the implementation.
- **`as unknown as` is required**, not a plain type annotation. `memo(forwardRef(...))` is not assignable to the two-parameter call signature, because `name`'s type is a deferred conditional. The single-parameter version on `master` did typecheck as a plain annotation, so this is a real loss of type-checking at the definition site: nothing verifies the cast matches the implementation. On both platforms this is now the only thing standing between the declared API and the implementation.

### Deriving the name union from the passed component

Two small type-level pieces:

```ts
// packages/common/src/types/IconComponent.ts
export type IconNameOf<IconComponentType> = IconComponentType extends (props: infer Props) => any
  ? Props extends { name: infer Name } ? Name : never
  : never;
```

`createIcon` returns `memo(forwardRef(...))` on web and `memo(...)` on mobile; both expose a call signature, so `(props: infer Props) => any` reads through the wrappers on both platforms.

`IconLike` is the minimal structural constraint, defined per platform next to `createIcon`:

```ts
export type IconLike = (props: {
  name: any;
  size?: IconProps['size'];
  active?: IconProps['active'];
  color?: IconProps['color'];
  /* ...the props CDS components actually drive the icon with */
}) => React.ReactNode;
```

`name` **has to be `any`**. Props are checked contravariantly, so requiring `name: string` rejects exactly the components we want to accept — a set-bound component from `createIcon` narrows `name` to its own union. The remaining entries are sourced from `IconProps` so the contract tracks `Icon` itself.

Honest limitation of this constraint: because function parameters are contravariant, `IconLike` can only guarantee the *shape of `name`*. A component that silently ignores `size`, `color`, or `active` still satisfies it, and CDS has no way to detect that. Any icon component not built with `createIcon()` is on the honor system for size and color fidelity.

## Backward compatibility

Fully additive. Both `IconButtonBaseProps`/`IconButtonProps` and `BannerBaseProps`/`BannerProps` gained a **defaulted** type parameter, so every existing reference (`ConfigResolver<IconButtonBaseProps>`, `Omit<IconButtonBaseProps, 'name'>` in `FeedCard`, `IconButtonProps<IconButtonDefaultElement>` in `InputIconButton`, `Partial<BannerProps>` in stories and `apps/docs`) still resolves. Omitting `IconComponent` reproduces today's types and behavior exactly — size derivation, variant colour resolution, accessibility props, and the loading state are untouched. Zero existing call sites changed.

## UI changes

None. Every existing call site renders byte-identically; the only new rendering path is the opt-in one, which is covered by tests.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

**Type-level tests are the heart of this POC.** Four new test files assert the three DX cases (plus a fourth: a custom name is rejected when no `IconComponent` is supplied), using the repo's existing `@ts-expect-error` idiom for the negative cases. Both packages' tsconfigs `include: ["src/**/*"]`, so these files are covered by `typecheck` — an `@ts-expect-error` that stops being an error fails the build as an unused directive. That is what makes the negative assertions load-bearing rather than decorative.

Each platform gets a self-contained fake icon set built via `createIcon()` (`src/icons/__fixtures__/fakeIconSet.ts`) with names chosen not to collide with any built-in name, so the POC takes on no dependency on `@cbhq/retail-icons`.

<details>
<summary>Verification that the negative assertions actually fail (not assumed)</summary>

I flipped case 3 from the invalid name to a valid one and confirmed the build breaks:

```
$ yarn nx run web:typecheck   # with name="fakeSatellite" instead of name="close"
Found 1 error.   # unused '@ts-expect-error' directive
```

And captured the real diagnostics with the suppressions removed, confirming the error lands on `name` with the union derived from the inferred `IconComponent` rather than on some unrelated prop:

```
error TS2322: Type '"close"' is not assignable to type '"alpha" | "beta"'.
  The expected type comes from property 'name' which is declared here on type
  'IntrinsicAttributes & WidgetProps<"button", NamedExoticComponent<Omit<IconProps<"alpha" | "beta">, "ref"> & RefAttributes<HTMLElement>>>'
```

</details>

### Testing instructions

```bash
# unit tests for the specific files changed
cd packages/web  && npx jest src/buttons/__tests__/IconButton.test.tsx \
  src/buttons/__tests__/IconButtonIconComponent.test.tsx \
  src/banner/__tests__/Banner.test.tsx src/banner/__tests__/BannerIconComponent.test.tsx
cd packages/mobile && npx jest src/buttons/__tests__/IconButton.test.tsx \
  src/buttons/__tests__/IconButtonIconComponent.test.tsx \
  src/banner/__tests__/Banner.test.tsx src/banner/__tests__/BannerIconComponent.test.tsx

# typecheck (this is where the negative type assertions are enforced)
yarn nx run-many --target=typecheck --projects=web,mobile,storybook,expo-app,docs

# full suites, to confirm nothing regressed
yarn nx run-many --target=test --projects=web,mobile
```

Results:

| Command | Result |
| --- | --- |
| Targeted web jest (4 files) | 4 suites / 51 tests passed |
| Targeted mobile jest (4 files) | 4 suites / 33 tests passed |
| `typecheck` on web, mobile, storybook, expo-app, docs | passed (`--skip-nx-cache`) |
| `test` on web + mobile (full) | 391 suites / 4,566 tests passed, 0 failed |
| `prettier --check` on changed files | clean |
| `eslint` on changed files | no new warnings |

`eslint` reports 2 pre-existing `internal/no-style-prop-css-overrides` warnings on `web/src/buttons/IconButton.tsx` and export-sort warnings on both `icons/index.ts` files; I confirmed all of these are present on `master` unchanged and left them alone.

## What this approach does *not* solve

Being specific, since this is the crux of the comparison:

- **It cannot reach icons CDS hardcodes internally.** `IconComponent` only reaches names the *consumer* supplies. Glyphs CDS chooses itself are unreachable: `Banner`'s dismiss `close` button (demonstrated by a passing test in this PR), disclosure carets, selection checkmarks, `Select`/`Dropdown` chevrons, `SearchInput`'s clear affordance. A consumer who swaps in their own icon set gets a component rendering **two icon fonts side by side**, which is arguably worse than not swapping at all. Reaching those would need a *separate* mechanism — and the natural one is a runtime provider, i.e. the adopted approach.
- **It puts a per-call-site burden on consumers.** Every single JSX site must pass `IconComponent={RetailIcon}`. For an app the size of Coinbase Retail that is hundreds of edits, it is easy to miss one, and there is no lint rule or type error to catch a site that was forgotten — a missed site silently falls back to the built-in set and either renders the wrong glyph or none. The adopted provider approach is one wrap at the app root.
- **No runtime guard.** The guarantee is purely compile-time. A mismatched name still constructs and renders at runtime (there is an explicit test documenting this). Anything crossing a JS boundary — server-driven UI, `any`-typed props, `React.cloneElement` — bypasses it entirely.
- **The two mechanisms don't compose for free.** `IconComponent` narrows names *down* per call site; the registry widens them *up* globally. Shipping both means two ways to do the same thing with different failure modes, and a consumer using both gets the union of names via the registry regardless of what `IconComponent` says.

## Comparison with the adopted provider approach

| | This POC (`IconComponent`) | Adopted (`IconGlyphSourceProvider` + `Cds.IconNameRegistry`) |
| --- | --- | --- |
| Type-safety scope | Per-call-site — a name is only valid where its glyph exists | Global — any registered name valid everywhere, incl. subtrees without the glyph |
| Consumer effort | Every call site | One provider wrap + one `declare global` block |
| Reaches CDS-internal icons | **No** | **Yes** — `Icon` consults the provider wherever it renders |
| Mixed-font risk within one component | **Yes** (see `Banner` dismiss) | No |
| CDS effort per component | Two type params, a hand-written call signature, an `as unknown as` cast, a props-type reindent | **Zero** — components are untouched |
| Runtime cost | None | One context read per `Icon` |
| Runtime safety | None (compile-time only) | Glyph resolution falls back through sources |

The honest summary: this POC's per-call-site precision is genuinely better *type safety*, and it is the only one of the two that can reject `name="close"` under a retail-only icon set. But it buys that precision at a cost that scales with **both** the number of CDS components and the number of consumer call sites, and it structurally cannot solve the internal-icon half of the problem. The adopted approach scales at O(1) for consumers and O(0) for CDS.

## What would it cost to roll out?

Beyond the two components here, **~12 components / ~21 files** still type an icon prop to the closed union:

- **Cross-platform (web + mobile):** `Button` (2 names), `Tag` (2), `Stepper` (6), `DefaultCarouselNavigation` (4), `CellMedia`, `DotSymbol`, `SectionHeader`, `IconCounterButton`, `SearchInput`
- **Platform-specific:** `NavBarIconButton` (mobile), `SegmentedControl` (web), `DefaultPaginationNavigationButton` (web)

Per component that means: a defaulted type parameter on both the `*BaseProps` and `*Props` types, a hand-written generic call signature (most are not polymorphic, so this is net-new), an `as unknown as` cast, an `IconLike` widening at the render site, and a props-type reindent that inflates the diff (see `mobile/src/banner/Banner.tsx` in this PR — ~40 lines of pure indentation churn).

Three complications the proposal doc does not currently anticipate, which I hit while implementing:

1. **Composite components have to thread the parameter through.** `DefaultCarouselNavigation` renders three `IconButton`s. To work it must accept `IconComponent`, derive all four of its own name props from it, *and* forward it to each child `IconButton` — the generic has to be plumbed at every layer of composition. Miss a layer and the child silently falls back to the built-in set. This is the cost that actually grows non-linearly, and it is why I'd expect a full rollout to be considerably more than "12 × the IconButton diff".
2. **`Stepper`'s names are nested inside an object prop** (`icon?: { defaultName?, activeName?, ... }`). Deriving those from a top-level `IconComponent` means making the nested object type generic too, which is materially uglier than the flat case and pushes the type parameter through a second layer.
3. **The definition-site cast loses type-checking.** Every one of the ~21 files would carry an unchecked `as unknown as`, so nothing verifies the declared public API matches the implementation. That's a meaningful erosion of safety in a library, applied uniformly, in service of adding safety at the call site.

## My assessment

Having implemented it: **the mechanism works exactly as the proposal hoped, and I'd still not ship it library-wide.**

The type derivation is genuinely elegant, survives `memo`/`forwardRef` with a known workaround, and delivers precisely the per-call-site guarantee the doc claims — case 3 in the sketch above really is a compile error, and I verified that with a deliberately-broken control run rather than by assuming. If per-call-site precision were the only axis, this would win.

But two things I only saw clearly after building it change the conclusion. First, **it solves less than half the actual problem.** A consumer who adopts it still can't touch the glyphs CDS chooses internally, so the end state is components rendering two icon fonts at once — the `Banner` dismiss test in this PR is that failure mode, passing and asserted. Fixing that requires a runtime injection mechanism, at which point you've built the adopted approach anyway and this becomes redundant surface area. Second, **the cost is in composition, not in the leaf components.** `IconButton` is the easy case and the one the proposal uses; `Banner` was already harder (no existing generic to extend), and `DefaultCarouselNavigation` shows the parameter has to be threaded through every layer with no type error when a layer is missed.

So my recommendation is to keep the adopted provider approach as the mechanism and **not** roll this out broadly. What I'd genuinely reconsider is a *narrow* version: `IconComponent` on the two or three components where consumers most often want an icon CDS doesn't have and where the component renders exactly one, consumer-supplied icon and nothing else — `IconButton` and `CellMedia` fit. There it's cheap, complete (no internal glyphs to strand), and composes with the provider rather than competing with it. Library-wide, the ledger doesn't balance.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/packages/common/src/types/IconComponent.ts
+++ b/packages/common/src/types/IconComponent.ts
@@ -1,0 +1,16 @@
+/**
+ * Derives the accepted `name` union from an icon component's props.
+ *
+ * Components that render an icon by name take an optional `IconComponent` prop;
+ * their `name` prop is typed as `IconNameOf<typeof IconComponent>` so it narrows
+ * to whatever icon set was passed, and to the built-in `IconName` when omitted.
+ *
+ * The `(props: infer Props) => any` pattern reads through the wrappers
+ * `createIcon` applies — `memo(forwardRef(...))` on web and `memo(...)` on
+ * mobile both expose a call signature, so the props are recoverable from either.
+ */
+export type IconNameOf<IconComponentType> = IconComponentType extends (props: infer Props) => any
+  ? Props extends { name: infer Name }
+    ? Name
+    : never
+  : never;
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -22,6 +22,7 @@ export * from './FallbackBaseProps';
 export * from './Grid';
 export * from './Helpers';
 export * from './IconButtonBaseProps';
+export * from './IconComponent';
 export * from './IconName';
 export * from './IconSize';
 export * from './IllustrationNames';
--- a/packages/mobile/src/banner/Banner.tsx
+++ b/packages/mobile/src/banner/Banner.tsx
@@ -3,12 +3,13 @@ import type { StyleProp, TextStyle, View, ViewStyle } from 'react-native';
 import type { ThemeVars } from '@coinbase/cds-common/core/theme';
 import { variants } from '@coinbase/cds-common/tokens/banner';
 import type { BannerStyleVariant, BannerVariant } from '@coinbase/cds-common/types/BannerBaseProps';
-import type { IconName } from '@coinbase/cds-common/types/IconName';
+import type { IconNameOf } from '@coinbase/cds-common/types/IconComponent';
 import type { SharedProps } from '@coinbase/cds-common/types/SharedProps';
 
 import { Collapsible } from '../collapsible/Collapsible';
 import { useComponentConfig } from '../hooks/useComponentConfig';
 import { useTheme } from '../hooks/useTheme';
+import type { IconLike } from '../icons/createIcon';
 import { Icon } from '../icons/Icon';
 import { Box } from '../layout/Box';
 import type { HStackProps } from '../layout/HStack';
@@ -18,11 +19,25 @@ import { Pressable } from '../system/Pressable';
 import { Link, type LinkProps } from '../typography/Link';
 import { Text } from '../typography/Text';
 
-export type BannerBaseProps = SharedProps & {
+export type BannerBaseProps<IconComponentType extends IconLike = typeof Icon> = SharedProps & {
   /** Sets the variant of the banner - which is responsible for foreground and background color assignment */
   variant: BannerVariant;
-  /** Name of icon to be shown in the banner */
-  startIcon: IconName;
+  /**
+   * Name of icon to be shown in the banner. Accepted values are derived from
+   * `IconComponent`, so they narrow to a custom icon set's names when one is
+   * passed and stay the built-in `IconName` otherwise.
+   */
+  startIcon: IconNameOf<IconComponentType>;
+  /**
+   * Component used to render `startIcon`. Pass an icon component built with
+   * `createIcon` to render icons from a set CDS does not ship; `startIcon` is then
+   * type-checked against that set instead of the built-in names.
+   *
+   * Note that this reaches only `startIcon`. The dismiss button's `close` icon is
+   * chosen by Banner itself and always renders from the built-in set.
+   * @default Icon
+   */
+  IconComponent?: IconComponentType;
   /** Whether the start icon is active */
   startIconActive?: boolean;
   /** Provide a CDS Link component to be used as a primary action. It will inherit colors depending on the provided variant */
@@ -74,43 +89,54 @@ export type BannerBaseProps = SharedProps & {
  * Note: docs docgen expects a `${ComponentName}Props` export (for Banner -> BannerProps)
  * to generate styles selector metadata used by the docs Styles tab.
  */
-export type BannerProps = BannerBaseProps &
-  Omit<HStackProps, 'children'> & {
-    /** Slot-level styles for Banner. */
-    styles?: {
-      /** Persistent outer wrapper around both dismissible and non-dismissible variants. */
-      root?: StyleProp<ViewStyle>;
-      /** Main content container (`HStack`) for banner body. */
-      content?: StyleProp<ViewStyle>;
-      /** Start icon wrapper. */
-      start?: StyleProp<ViewStyle>;
-      /** Right-side body wrapper containing middle content and actions. */
-      body?: StyleProp<ViewStyle>;
-      /** Middle content wrapper containing title/message/label region. */
-      middle?: StyleProp<ViewStyle>;
-      /**
-       * Label text style.
-       * Applies only when `label` is a string rendered by Banner.
-       * If `label` is a custom node, style that node directly.
-       */
-      label?: StyleProp<TextStyle>;
-      /**
-       * Actions row style.
-       * Applies only when at least one action (`primaryAction` or `secondaryAction`) is rendered.
-       */
-      actions?: StyleProp<ViewStyle>;
-      /**
-       * Dismiss button wrapper style.
-       * Applies only when `showDismiss` is true.
-       */
-      dismiss?: StyleProp<ViewStyle>;
+export type BannerProps<IconComponentType extends IconLike = typeof Icon> =
+  BannerBaseProps<IconComponentType> &
+    Omit<HStackProps, 'children'> & {
+      /** Slot-level styles for Banner. */
+      styles?: {
+        /** Persistent outer wrapper around both dismissible and non-dismissible variants. */
+        root?: StyleProp<ViewStyle>;
+        /** Main content container (`HStack`) for banner body. */
+        content?: StyleProp<ViewStyle>;
+        /** Start icon wrapper. */
+        start?: StyleProp<ViewStyle>;
+        /** Right-side body wrapper containing middle content and actions. */
+        body?: StyleProp<ViewStyle>;
+        /** Middle content wrapper containing title/message/label region. */
+        middle?: StyleProp<ViewStyle>;
+        /**
+         * Label text style.
+         * Applies only when `label` is a string rendered by Banner.
+         * If `label` is a custom node, style that node directly.
+         */
+        label?: StyleProp<TextStyle>;
+        /**
+         * Actions row style.
+         * Applies only when at least one action (`primaryAction` or `secondaryAction`) is rendered.
+         */
+        actions?: StyleProp<ViewStyle>;
+        /**
+         * Dismiss button wrapper style.
+         * Applies only when `showDismiss` is true.
+         */
+        dismiss?: StyleProp<ViewStyle>;
+      };
     };
-  };
 /**
  * @deprecated Use `BannerProps` instead. This will be removed in a future major release.
  * @deprecationExpectedRemoval v10
  */
 export type MobileBannerProps = BannerProps;
+
+/**
+ * Unlike `IconButton`, `Banner` is not polymorphic, so it had no generic call
+ * signature to extend — one has to be introduced purely to carry the icon set
+ * through `memo`, which erases type parameters. This is the per-component cost of
+ * the pattern for the majority of CDS components.
+ */
+type BannerComponent = (<IconComponentType extends IconLike = typeof Icon>(
+  props: BannerProps<IconComponentType> & { ref?: React.Ref<View> },
+) => React.ReactNode) & { displayName?: string };
 
 export const Banner = memo(function Banner({
   ref: forwardedRef,
@@ -122,6 +148,7 @@ export const Banner = memo(function Banner({
   const {
     variant,
     startIcon,
+    IconComponent,
     startIconActive,
     onClose,
     primaryAction,
@@ -149,6 +176,7 @@ export const Banner = memo(function Banner({
   } = mergedProps;
   const [isCollapsed, setIsCollapsed] = useState(false);
   const theme = useTheme();
+  const ResolvedStartIcon: IconLike = IconComponent ?? Icon;
 
   // Events
   const handleOnDismiss = useCallback(() => {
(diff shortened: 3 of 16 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
🟡 See below
</td>
<td></td>
</tr>
</table>

<h2>
🟡
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
🟡
<code>0/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #863, opened 2026-08-27
fix: cut mobile visreg run time to stop 30m poll timeouts

cb-ekuersch:
## What changed? Why?

Mobile visreg jobs were failing consistently. The fix reduces real device time rather than just raising the ceiling, though it does both.

**Flow-level latency (the bulk of the win).** `waitForAnimationToEnd` defaults to 15s and only returns early once the screen is pixel-stable. Routes that animate indefinitely — the charts, `Carousel`, `Coachmark`, `SlideButton`, `Dot` — never stabilize, so each call burned the full 15s *and still* screenshotted mid-animation. The suite made 106 unbounded calls across 47 routes. Both sub-flows now pass explicit timeouts; static routes are unaffected because they settle well inside the cap.

**Overlay dismissal.** The flow tried tapping `Cancel`, then `Close`, both `optional`. Maestro's optional-element lookup is a hardcoded 7s that cannot be tuned per command, so all six overlay routes paid it whichever way it resolved. `overlayRoutes` is now a map from route to its exact dismiss label, so there is one non-optional tap. This also closes a silent failure mode: previously a renamed control left the overlay open and polluted the *next* route's screenshot.

**Diagnosability.** Queue time and execution time now have separate budgets, so a saturated BrowserStack account is distinguishable from a slower suite — with one combined budget both presented identically as "timed out at 30 minutes". Every run prints a `Timing: queued Xm, ran Ym, downloaded Zm` line, and all budgets are env-overridable so they can be tuned from CI without a code change.

**Two latent hangs and a debuggability gap found along the way:**

- `bsFetch` had no request timeout. Node's fetch has none by default and the poll loop only checked its deadline *between* requests, so a stalled connection would have outlived the build budget entirely and run until GitHub killed the job at 6h. Requests are now bounded, GETs retry on transient failures (a run issues 100+ poll requests; a single 503 twenty-five minutes in used to discard the whole build), and jobs get a 75m backstop.
- The Percy step discarded its own output on failure. Under `bash -e`, `OUTPUT=$(...)` aborts the step before the `echo`, which is exactly why failed runs showed an empty Percy error and skipped the PR comment.
- Screenshot artifacts now download 8-at-a-time instead of serially, and uploads get their own 10m budget (the APK is 59MB, so the default 60s cap would have been a regression).

Failures were the `Run visreg on BrowserStack` step hitting `POLL_TIMEOUT_MS = 30 * 60 * 1000` in `browserstack.mjs` with the build still reporting `running` — not a GitHub-level kill (no `timeout-minutes` was set, so the default was 6h).

Every failure landed at a uniform ~30 minutes while passing runs took 15–27 minutes, which identified this as a margin problem rather than a hang: the suite had grown until it sat on the ceiling. Across four sampled failures iOS and Android alternated as the victim while the other platform passed in the same run.

## UI changes

None. CI tooling and Maestro flow configuration only; no component source is touched.

The capped animation waits do not change what is captured on any route that settles. Routes that animate indefinitely were already being screenshotted mid-animation, since the old 15s wait never observed a static frame either.

### How has it been tested?

- [ ] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

Verified locally without a device (this package has no Jest target):

- Flow generation produces 47 routes — 41 standard, 6 overlay, all 6 carrying a `DISMISS_LABEL`.
- All three flow YAMLs parse.
- The new poll/retry logic was smoke-tested against a stubbed BrowserStack API: happy path timing, queue-specific vs run-specific timeout errors, transient 503 recovery, and immediate failure on a non-retryable 401. That test caught a real bug in the first pass — the non-retryable `throw` sat inside its own `try`, so a 401 was retried three times instead of failing immediately.
- `yarn nx format:check` is clean.

Dismiss labels were confirmed against the story sources rather than guessed: `AlertBasic`, `DrawerLeft`, `DrawerTop`, `ModalBasic` and `StickyFooter` use `Cancel`; `TrayBasic` uses `Close`.

### Testing instructions

This is unvalidated against a real device until CI runs. The `Timing:` line in the job log is the thing to check — it shows how much the animation caps actually bought and whether any remaining slowness is queue contention or device execution.

1. Confirm both `Visreg iOS` and `Visreg Android` pass, and compare the `ran` figure against the 15–27 minute baseline.
2. Confirm the Percy build receives snapshots from both platforms and the Percy link is commented on this PR.
3. Spot-check the six overlay snapshots (`AlertBasic`, `DrawerLeft`, `DrawerTop`, `ModalBasic`, `StickyFooter`, `TrayBasic`) for the overlay being open, and check that the route captured immediately after each one is not showing a leftover overlay.
4. Animation-heavy routes (`AreaChart`, `BarChart`, `CartesianChart`, `Carousel`, `Coachmark`, `SlideButton`, `Dot`) were never deterministic; if any now diff, that is the pre-existing instability surfacing at a different frame rather than a regression from this change.

Not applicable — no files under `packages/illustrations/**` or `packages/icons/**`.

## Notes for reviewers

One thing deliberately left out: a per-platform result cache, so an Android timeout would not force iOS to re-run. With `PERCY_PARALLEL_TOTAL: 2`, skipping a platform strands the Percy build waiting on a shard that never arrives, so the check would hang rather than fail. It needs a dynamic parallel-total and would produce single-platform Percy builds for reviewers — worth its own PR.

`.github/workflows/visreg-web.yml` has the same `set -e` output-swallowing bug in its Percy step. Left alone to keep this scoped to mobile.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false

### DIFF
--- a/.github/workflows/visreg-mobile.yml
+++ b/.github/workflows/visreg-mobile.yml
@@ -42,6 +42,10 @@ jobs:
     needs: [check]
     if: needs.check.outputs.should_run == 'true'
     runs-on: ubuntu-latest
+    # Backstop only. The run script enforces its own queue/execution budgets and
+    # exits with a diagnosis; this exists so a hang that escapes those budgets
+    # fails in an hour instead of burning the 6h GitHub default.
+    timeout-minutes: 75
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
@@ -90,8 +94,13 @@ jobs:
         id: percy-upload
         if: always()
         run: |
+          # `set +e` around the capture: the step runs under `bash -e`, where a
+          # failing command substitution aborts before the echo below, hiding
+          # the Percy output that explains the failure.
+          set +e
           OUTPUT=$(yarn nx run mobile-visreg:upload 2>&1)
           EXIT_CODE=$?
+          set -e
           echo "$OUTPUT"
           PERCY_URL=$(echo "$OUTPUT" | grep -oE 'https://percy\.io[^[:space:]]+' | head -1)
           echo "percy_url=$PERCY_URL" >> "$GITHUB_OUTPUT"
@@ -115,6 +124,8 @@ jobs:
     needs: [check]
     if: needs.check.outputs.should_run == 'true'
     runs-on: ubuntu-latest
+    # See the iOS job — backstop only.
+    timeout-minutes: 75
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@ec9f2d5744a09debf3a187a3f4f675c53b671911 # v2.13.0
@@ -178,8 +189,13 @@ jobs:
         id: percy-upload
         if: always()
         run: |
+          # `set +e` around the capture: the step runs under `bash -e`, where a
+          # failing command substitution aborts before the echo below, hiding
+          # the Percy output that explains the failure.
+          set +e
           OUTPUT=$(yarn nx run mobile-visreg:upload 2>&1)
           EXIT_CODE=$?
+          set -e
           echo "$OUTPUT"
           PERCY_URL=$(echo "$OUTPUT" | grep -oE 'https://percy\.io[^[:space:]]+' | head -1)
           echo "percy_url=$PERCY_URL" >> "$GITHUB_OUTPUT"
--- a/packages/mobile-visreg/README.md
+++ b/packages/mobile-visreg/README.md
@@ -13,9 +13,9 @@ Visual regression testing for CDS mobile components. Runs Maestro flows on [Brow
 ```
 packages/mobile-visreg/
   config/
-    enabled-routes.mjs          # Explicit opt-in list of routes + overlay route set
+    enabled-routes.mjs          # Explicit opt-in list of routes + overlay route → dismiss label
   src/
-    config.mjs                  # Re-exports enabled routes + default settings
+    config.mjs                  # Re-exports enabled routes + timing budgets
     generate-flows.mjs          # Generates flows/capture-all.yaml from the route list
     browserstack.mjs            # BrowserStack App Automate REST API client
     browserstack-run.mjs        # Orchestrator CLI — uploads app+flows, triggers build, downloads screenshots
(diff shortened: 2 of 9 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

adrienzheng-cb (approved):

PR #866, opened 2026-08-28
chore: replace custom versioning scripts with nx release version plans

adrienzheng-cb:
## What changed? Why?

Replaces our hand-rolled versioning and changelog tooling with Nx's built-in release process, matching what `frontend/cds` recently adopted.

Deleted: `tools/bumpVersion.mjs`, `tools/validateCDSVersions.mjs`, `tools/ci/getProjectsNeedingVersion.mjs`, `tools/ci/validators/validateVersioned.mjs`, `libs/codegen/src/release/updatePkgsForGenericBump.mjs` and the `update-packages-generic-bump` target, plus their specs — about 1,300 lines.

Added a `release` block to `nx.json` with two groups covering all 12 published packages:

- **`cds`** (fixed): `web`, `mobile`, `common`, `mcp-server` — keeps them in lockstep, which is what `validateCDSVersions.mjs` used to enforce. Because Nx versions the whole group from one plan, this also ends the `This is an artificial version bump with no new change.` changelog entries that the old generic-bump script produced.
- **`standalone`** (independent): the other eight packages, tagged `{projectName}@{version}`.

The new `cds-ios` and `cds-android` packages are `private: true`, so they stay outside `nx release` — consistent with their existing release docs.

### Developer flow

`yarn nx release plan` writes a version plan to `.nx/version-plans/`. A maintainer applies it with `yarn release` (`nx release --skip-publish`), which bumps versions, writes changelogs, and deletes the consumed plan. Both steps can happen in the same PR. Full write-up in `docs/release.md`.

There is deliberately no automated release workflow. CI cannot push to `master`: the ruleset requires a reviewed PR, signed commits, and linear history with no bypass actors, and a `GITHUB_TOKEN`-opened PR does not trigger the required status checks. So releasing stays a local, human-run step.

`publish.yml` is unchanged and still owns publishing. `yarn release` runs with `--skip-publish`, so merging a version bump to `master` is what triggers the npm publish, exactly as before.

### CI gate

`validateVersionPlans.mjs` replaces `validateVersioned.mjs` in the `Validate` job. It passes an affected publishable project if **either** a pending plan names it (or its fixed group) **or** its `package.json` version already moved in the branch. That second condition is what makes releasing in the same PR possible; `nx release plan:check` can only express the first.

### Orphan cleanup (second commit)

Removing the old scripts left three things stranded:

- `tools/ci/isReleaseBranch.mjs` and its spec — only `bumpVersion.mjs` imported it, and `publish.yml` reimplements the same `release-N.x` → `vN-lts` logic inline in bash.
- `tools/generateUpdates.mjs` — an unreferenced biweekly Slack digest that scraped changelog headings for the old `## 9.22.1 ((8/25/2026, 02:10 PM PST))` format, which Nx no longer emits.
- The `bump-version` and `changelog` scripts. `bump-version` was an exact duplicate of `changelog`, and `changelog` was itself misleading — it wrote a version plan, not a changelog. Callers now invoke `yarn nx release plan` directly.

`yarn release` is deliberately kept, since it is not a pure alias: it supplies `--skip-publish`, which cannot be set in `nx.json`, and without it `nx release --yes` would publish to npm from a local machine.

### Notable details

- `version.useLegacyVersioning: false` is required. The Nx 20.8.2 default (V1) crashes with `Version data key "common" already exists in version data` on our cross-group dependencies and silently ignores `updateDependents: "never"`.
- `conventionalCommits: false` is intentionally **omitted** rather than set — Nx 20.8.2 throws `Cannot convert undefined or null to object` when it is explicit under V2, and `false` is already the default.
- Changelog preambles moved to a footer in 11 of the 12 files, because Nx prepends new entries to the top. `packages/illustrations/CHANGELOG.md` is deliberately left as-is: the illustrations Figma sync lives in the internal `frontend/cds` repo and splices its entry in at the `<!-- template-start -->` marker, so removing that marker would make the sync silently write nothing. It should get the same treatment once that sync is ported here or updated to write a version plan (see `docs/release.md`).
- `.gitignore` now uses `**/.nx/*` with `!/.nx/version-plans` so plan files are tracked while cache and daemon state stay ignored.
- The Figma icon sync now writes a version plan (major if icons were renamed or deleted, otherwise minor) instead of splicing into `CHANGELOG.md` at the `<!-- template-start -->` marker.

## UI changes

None — tooling and docs only.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

### Testing instructions

`yarn nx run tools:test` passes (29 tests, including the new `getVersionPlanCoverage.spec.mjs` covering fixed-group resolution, already-bumped versions, and new packages).

End-to-end verified by writing a plan naming both a fixed-group and an independent package, running `yarn nx release --skip-publish --dry-run`, and confirming the four `cds` packages bump in lockstep, `icons` bumps independently, both changelogs render correctly, and the plan is consumed. The new validator was also exercised against a modified source file to confirm it fails without a plan and passes with one.

Note that the validator reports "No changes within packages" for this PR, which is correct — `getAffectedPackages` keys off `BUMP_REGEX` (`src/`, `package.json`, and similar) and this PR touches only scripts, changelogs, and config. The first feature PR after this merges will be the real exercise of the new gate.

Not applicable — no icon or illustration assets change. `packages/icons/scripts/sync-icons/` changes only affect how the sync script records its output.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

tools/ci/validators/validateVersionPlans.mjs
```diff
@@ -0,0 +1,48 @@
+import {
```
cb-ekuersch, line 1: nit: we can probably scrap the validators folder
also is everything in this"ci" folder truley only for ci? Might need a more accurate name for clarity

packages/web/CHANGELOG.md
```diff
@@ -1342,3 +1332,7 @@ This is an artificial version bump with no new change.
 ## 8.12.0 (9/18/2025 PST)
 
 - Prepare for open source release.
+
+---
+
+<sub>Release entries above are generated by `nx release` from version plans. See the [NPM registry](https://www.npmjs.com/package/@coinbase/cds-web) for published versions. `@coinbase/cds-web` adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).</sub>
```
cb-ekuersch, line 1344: nit: should we include the --skik publish flag in this? Or was this auto-genearted?

packages/icons/scripts/sync-icons/generateVersionPlan.ts
```diff
@@ -0,0 +1,43 @@
+import type { IconSyncResults } from './index';
+
+/** Renaming or deleting an icon breaks consumers, so it forces a major bump. See DOCS.md. */
+const isBreaking = (syncResults: IconSyncResults) =>
```
cb-ekuersch, line 4: doesn't illustrations need this too?

should this be unit tested?

packages/web/CHANGELOG.md
adrienzheng-cb: Hand-written, not auto-generated — it's part of this PR. Nx never touches it: it blind-prepends new entries above everything (`changelogContents = \`\${contents}\n\n\${changelogContents}\``), which is exactly why the preamble had to move down here. Left at the top, the first release would have inserted `## 9.22.2` above the `# @coinbase/cds-web` title.

On `--skip-publish`, I'd leave it out. That sentence describes how the *entries* are produced, and the flag has no bearing on changelog content — it only controls whether `npm publish` runs afterward. This footer renders on npm and GitHub for consumers of the package, so a maintainer-only flag would be noise. It is documented, in `docs/release.md` and `AGENTS.md`, which is where someone cutting a release would look.

tools/ci/validators/validateVersionPlans.mjs
adrienzheng-cb: You're right about `ci/`, and I can back it with a concrete case: `shouldRunVisreg.mjs` is invoked from `apps/storybook/package.json` (the `should-run-visreg` script) and from `packages/mobile-visreg/project.json`, both of which run locally. So the folder name is genuinely inaccurate today.

On `validators/` I'd push back slightly — it predates this PR (`validateIconSvgMap`, `validateLockfile`, `validatePatches` were already there) and I followed the existing convention. All four are invoked only from `ci.yml`, so that subfolder really is CI-only even though its parent isn't.

Either change is a pure rename touching `ci.yml`, `visreg-check.yml`, `pr.yml`, `apps/storybook/package.json`, `packages/mobile-visreg/project.json`, and the test imports — no behaviour change, but it'd bury the versioning diff in this PR. Happy to do it as a follow-up if you want; just say which shape you'd prefer (`tools/scripts/`, flatten `validators/` up, or both).

packages/icons/scripts/sync-icons/generateVersionPlan.ts
adrienzheng-cb: You were right, and my earlier reply on this thread was wrong.

<details>
<summary>That earlier reply is quoted here for the record (it was lost while I was annotating it)</summary>

> Illustrations doesn't need one. `generateVersionPlan.ts` exists because `sync-icons` pulls from Figma and used to write straight into `CHANGELOG.md` at the `<!-- template-start -->` marker, so it had to be taught the new format. Illustrations has no equivalent — its only scripts are `generateStories.ts` and `writeStories.ts`, and neither it nor `libs/codegen` ever wrote a changelog. Illustration releases already go through the normal by-hand version plan, same as any other package.

</details>

Illustrations does have a Figma sync, and it does write the changelog. It just isn't in this repo, which is why my search of `packages/illustrations/scripts` turned up nothing. It lives in the internal `frontend/cds` repo as `libs/illustration-tasks`, and its task config writes directly into a checkout of this one:

```ts
// libs/illustration-tasks/src/scripts/illustrations/task.ts
changelogFile: path.resolve(CDS_OSS_ROOT, "packages/illustrations/CHANGELOG.md"),
```

It inserts at exactly the marker this PR deleted:

```ts
// libs/illustration-tasks/src/tools/Changelog.ts
const insertionPoint = this.content.includes(UNRELEASED_HEADER)
  ? UNRELEASED_HEADER
  : TEMPLATE_START;
...
this.content = this.content.replace(insertionPoint, `${insertionPoint}\n\n${newContent}`);
```

That file has no `## Unreleased`, so it falls through to `<!-- template-start -->`. With the marker removed, `String.replace` finds nothing, returns the string unchanged, and throws nothing — the next illustrations sync would have written no changelog entry and still reported success. Silent, which is the bad kind.

Fixed in 4640556ef: `packages/illustrations/CHANGELOG.md` is restored to its previous shape so the sync keeps working, and `docs/release.md` now records why that one file does not get the footer treatment the other 11 do.

Longer term it should get a `generateVersionPlan.ts` exactly as you suggested, either by porting `illustration-tasks` into this repo or by changing the internal tool to write a version plan. The port looks very doable — `@cds/figma-api` already exposes everything it imports, including the Variables API the color styles need, and the only internal dependencies left are a handful of small helpers. Both are follow-ups; this PR just stops it from breaking.

On the test: good call, added in 990fb88e5. It pins the bump-type rule, which is the actual logic here — a rename or delete is breaking (consumers reference icons by name) so it forces `major`, while additions and updates stay `minor`. The `icons` project had no test target at all, so that commit also adds the same minimal jest wiring `eslint-plugin-cds` uses.

Separately, `packages/illustrations/project.json` has a target literally named `release` that just runs `generate-stories`. It does not collide functionally (`nx run illustrations:release` and `nx release` are different namespaces), but the name is confusing now. Worth renaming separately.

adrienzheng-cb: added unit test. Leaving illustrations out for now, since its sync script is not in the same repo.

cb-ekuersch (approved):

PR #870, opened 2026-09-01
feat(web): add typeahead keyboard selection to Select (Alpha)

adrienzheng-cb:
## What changed? Why?

Adds native `<select>`-style typeahead keyboard selection to the web `Select (Alpha)`:

- A short-lived, multi-character search buffer that resets ~500ms after the last keystroke.
- Repeated-key cycling: pressing the same letter repeatedly cycles through every option that starts with it, matching native `<select>` semantics.
- Works both when the listbox is closed (opens it, then focuses the match) and while it is open, via a window-level `keydown` listener so keystrokes are handled consistently even after focus moves into the portaled dropdown.
- Buffer construction and match resolution are extracted into a pure `typeahead.ts` module (`isPrintableTypeaheadKey`, `normalizeOptionText`, `getTypeaheadMatchIndex`) with focused unit tests.

The previous implementation only matched the first letter of an option and only at open time, so users could not type multiple characters or cycle between same-prefixed options.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

### Testing instructions

Focus a web `Select (Alpha)` and type characters: with the listbox closed, typing opens it and focuses the first match; with it open, continue typing to refine the prefix match, or repeat a single letter to cycle through same-prefixed options. See `typeahead.test.ts` for the pure matching logic and the added `Select.test.tsx` cases for integration behavior.

Linear: https://linear.app/coinbase/issue/CDS-2505

## Change management

type=routine
risk=low
impact=sev5

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/web/src/alpha/select/typeahead.ts
```diff
@@ -0,0 +1,45 @@
...
+const printableTypeaheadKeyRegex = /^[a-z0-9]$/i;
+
+export function isPrintableTypeaheadKey(key: string): boolean {
+  return printableTypeaheadKeyRegex.test(key);
+}
+
+type TypeaheadKeyEvent = Pick<KeyboardEvent, 'key' | 'ctrlKey' | 'metaKey' | 'altKey'>;
+
+// Bare printable key, no modifier (shared by closed and open paths).
+export function isTypeaheadKeyEvent(event: TypeaheadKeyEvent): boolean {
```
cb-ekuersch, line 13: nit: why not make this KeyboardEvent?

packages/web/src/alpha/select/Select.tsx
```diff
@@ -163,42 +168,103 @@ const SelectBase = memo(
...
+        readOnly,
+        refs.reference,
+        refs.floating,
+        appendToTypeaheadBuffer,
+        focusTypeaheadMatch,
+      ]);
+
+      useEffect(
+        () => () => {
+          if (typeaheadResetTimeoutRef.current) clearTimeout(typeaheadResetTimeoutRef.current);
```
cb-ekuersch: we should clear this in the cleanup of the effect that creates the timeout

cb-ekuersch: i see that the timeout is crated in a user interaction so i dont think this effect just to do cleanup in even necessary

packages/web/src/alpha/select/Select.tsx
```diff
@@ -163,42 +168,103 @@ const SelectBase = memo(
...
+          appendToTypeaheadBuffer(event.key);
+          pendingTypeaheadRef.current = true;
+          setOpen(true);
         },
-        [disabled, readOnly, open, setOpen],
+        [disabled, readOnly, open, setOpen, appendToTypeaheadBuffer],
       );
 
       useEffect(() => {
-        if (!open || !pendingTypeAheadKeyRef.current) return;
```
cb-ekuersch, line 183: can we label what these different effects are for? Maybe there is value in giving them names and pulling them into custom hooks just for oganization's sake (e.g. `useFocusInPortaledDropdown` etc.)

cb-ekuersch (commented):
There is a lot going on here - i think it may help to review together. We should potentailly consider the possibility of abstracting this entire typeahead behavior into a single hook that can be called in the component render instead of inlining all this complex react logc in an already compex/long component render body

packages/web/src/alpha/select/useTypeahead.ts
```diff
@@ -0,0 +1,45 @@
...
+const printableTypeaheadKeyRegex = /^[a-z0-9]$/i;
+
+export function isPrintableTypeaheadKey(key: string): boolean {
+  return printableTypeaheadKeyRegex.test(key);
+}
+
+type TypeaheadKeyEvent = Pick<KeyboardEvent, 'key' | 'ctrlKey' | 'metaKey' | 'altKey'>;
+
+// Bare printable key, no modifier (shared by closed and open paths).
+export function isTypeaheadKeyEvent(event: TypeaheadKeyEvent): boolean {
```
adrienzheng-cb: Kept it as a structural `Pick<KeyboardEvent, ...>` on purpose and added a comment explaining why: the same predicate runs against both React's synthetic event (the control's `onKeyDown`) and the native `KeyboardEvent` (the window listener). A plain `KeyboardEvent` type would reject the synthetic event at the control call site, so the structural type is what lets both paths share `isTypeaheadKeyEvent`.

packages/web/src/alpha/select/Select.tsx
adrienzheng-cb: Done — I removed the standalone cleanup-only effect and folded the `clearTimeout` into the return of the window-listener effect (the effect that actually drives the open typeahead interaction). A reset timeout is only ever pending as part of an open interaction, so that effect's cleanup clears it both on the open→closed transition and on unmount. See the reply on your follow-up comment for the reasoning about why the timer is safe.

adrienzheng-cb: Agreed — the dedicated cleanup-only effect is gone. Two things worth noting: (1) the reset timeout's callback only assigns `""` to a ref (`typeaheadBufferRef`), so it never calls setState or focus; a late fire on an unmounted tree is a harmless no-op, not a stray-timer bug. (2) Even so, I kept an explicit clear by folding it into the window-listener effect's cleanup, which runs on the open→closed transition and on unmount — so no timer lingers. Net: no separate effect just for cleanup, and no stray timer.

packages/web/src/alpha/select/Select.tsx
adrienzheng-cb: This is largely addressed: all the typeahead effects were extracted out of `Select.tsx` into a dedicated `useTypeahead` hook (`useTypeahead.ts`), which is the "pull them into a custom hook for organization" direction you suggested. Inside the hook I also added terse labels — the type-to-open focus effect and the window-listener effect each have a one-line comment describing their purpose. I intentionally kept it as a single cohesive hook rather than splitting into several micro-hooks (e.g. `useFocusInPortaledDropdown`), since the effects share the buffer/timeout/focus refs and splitting would add indirection without real clarity. Happy to split further if you feel strongly.

packages/web/src/alpha/select/useTypeahead.ts
adrienzheng-cb: Switched this to the explicit `React.KeyboardEvent | KeyboardEvent` union for readability — it accepts both React's synthetic event from the control `onKeyDown` and the native event from the window `keydown` listener, without the structural `Pick`.

cb-ekuersch (dismissed):

cb-ekuersch (approved):

PR #871, opened 2026-09-02
feat: port the illustration Figma sync into this repo

adrienzheng-cb:
## What changed? Why?

The illustration sync lived outside this repo and wrote its generated assets and changelog entry into a *separate clone* of this one. Releasing illustrations therefore meant checking out two copies of the repo into specifically-named directories, running the sync from one, then hand-writing a changelog entry in the other. It also depended on a `<!-- template-start -->` marker in `packages/illustrations/CHANGELOG.md` to splice its entry in — the one thing blocking that file from getting the footer treatment the nx release migration (#866) gave every other package.

It now lives at `packages/illustrations/scripts/sync-illustrations/`, shaped like the icon sync next door, and writes an **nx version plan** rather than editing `CHANGELOG.md`. A sync releases through the same path as any other change.

```sh
yarn nx run illustrations:sync-illustrations
```

The script creates an `illustrations/YYYY-MM-DD` branch from `origin/master`, syncs from Figma, regenerates the docsite stories, writes `.nx/version-plans/illustrations-YYYY-MM-DD.md`, then commits and pushes. The runbook is in [`packages/illustrations/DOCS.md`](packages/illustrations/DOCS.md), mirroring `packages/icons/DOCS.md`.

### What came across, and what didn't

Only the **30 files reachable from the sync entry point**. The original was 60 files / ~8.4k lines, but roughly half of it was icon-font code (`helpers/font/*`, `oldGlyphIcons`, `uiIconExceptions`, `ComponentSet*`, `Validator`, `validators/*`) that the illustration pipeline never called. The port is ~3.6k lines.

Three dependencies had to be resolved:

| Was | Now |
| --- | --- |
| a shared Figma client | `@cds/figma-api` — already exports everything needed, including the Variables API |
| a shared script-utils package | vendored under `utils/` |
| a task-runner package | only supplied a `Task` type; replaced by the local `SyncTask` shape |

Vendoring rather than adding a shared lib is deliberate: `sync-icons` is self-contained too, importing only `@cds/figma-api`, npm packages, and its own siblings. `@nx/devkit` and `dotenv` were also dropped — the former was an undeclared transitive dependency used for two log lines, and the latter loaded a root `.env` this repo doesn't use (the token is exported into the environment, as the icons runbook already documents).

The class-based internals (`Manifest`, `Component`, `ColorStyles`) are kept as-is. This is a codegen pipeline where subtle breakage is expensive, so the outer shape moved and the working logic didn't. The pieces that genuinely had to change are the config (two repo roots collapsed to one), the git orchestration (now `ensureCleanBranch`, matching icons), and the changelog writer (now `generateVersionPlan.ts`).

The captured Figma fixture is stripped of the `user` block on each component — author name, id, and avatar URL — which the sync never reads and which has no business in a public repo.

### Also in here

- **Fixes the icon version plan.** It used bold labels (`**⭐️ Added (2)**`) for its sections. nx strips the blank lines out of a plan body when nesting it under the changelog entry, which turned every label following a list into a lazy continuation of that list's last item — rendering as `send ⭐️ Updated (1)` on one line. Both generators now use headings, which survive the collapsing and match what these changelogs have always looked like. This was my bug from #866.
- `packages/illustrations/CHANGELOG.md` gets the footer treatment, and the exception documented in `docs/release.md` is gone.
- **Includes a patch version plan.** The sync devDependencies go in `packages/illustrations/package.json`, matching how `packages/icons` declares its own sync deps, which the plan check counts as publishable. Requiring a plan does not cut a release — plans accumulate until someone runs `yarn release`, and nx keeps the largest bump among them — so this just records the change in the changelog whenever illustrations next ships.
- `scripts/` is now typechecked via `tsconfig.scripts.json`, wired into the existing `typecheck` target. `sync-icons` deserves the same and doesn't have it; worth a follow-up.

The 30 generated files under `src/__generated__` still name the old sync command in their `Generated from` header. Correcting it touches `packages/illustrations/src/`, which trips the designer sign-off gate below, and those boxes attest to a visreg and naming review that a comment edit has no need for. The next real sync rewrites these files and fixes the header for free.

### How has it been tested?

- [x] Unit tests

### Testing instructions

`yarn nx run illustrations:test` — 37 tests across 6 suites. Five suites are ported (`fetchIllustrationLibrary`, `getColorModeAndName`, `getPaintFromNode`, `getSize`, `parseName`); `generateVersionPlan.test.ts` is new and covers bump selection and the rendered markdown.

`yarn nx run illustrations:typecheck` and `illustrations:lint` are both clean.

The release integration was verified end to end by generating a plan from the real generator and dry-running nx release against it:

- additions/updates only → `4.48.0` → `4.49.0` (minor)
- with a rename → `5.0.0` under `⚠️ Breaking Changes` (major)

In both cases the entry rendered correctly and prepended above the existing `4.48.0` entry.

**Not exercised: the Figma calls themselves.** Those need the team token, so a real sync is the remaining verification. The script was smoke-tested to the point of its clean-tree guard, confirming config resolution, module loading, and the `figma-api:build` dependency.

Two things reviewers should know about a first real run:

1. **PNG output.** `sharp` is pinned at `^0.35.4` rather than the original `^0.31.2`, which predates Node 22 and has no usable prebuilt binaries here. Rasterization can differ subtly between libvips versions, so expect PNG bytes to change for any illustration the sync touches. Only changed illustrations are rewritten, so the blast radius is bounded to what would be rewritten anyway.
2. **Stale troubleshooting.** The old runbook described static color manifests checked into the sync's own directory. Those files don't exist any more — colors come from the Variables API. `DOCS.md` documents the actual behavior.

> Nothing under `packages/illustrations/src/` or `packages/icons/src/` changes in this PR — it moves tooling only. No assets or names are added, renamed, or removed, so there is nothing here for visreg or naming review to look at.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/illustrations/package.json
```diff
@@ -32,7 +32,13 @@
     "@babel/preset-env": "^7.28.0",
     "@babel/preset-react": "^7.28.5",
     "@babel/preset-typescript": "^7.27.1",
+    "@cds/figma-api": "workspace:^",
```
cb-ekuersch, line 35: check this

packages/illustrations/project.json
```diff
@@ -25,6 +25,27 @@
     "lint": {
       "executor": "@nx/eslint:lint"
     },
+    "test": {
+      "executor": "@nx/jest:jest",
```
cb-ekuersch, line 29: do we use the jest executor on our other packages?

.nx/version-plans/illustrations-sync-tooling.md
```diff
@@ -0,0 +1,5 @@
+---
+illustrations: patch
+---
+
+Add the `illustrations:sync-illustrations` target, which syncs illustrations from Figma and writes a version plan describing what changed. Adds sync-only devDependencies; no runtime change for consumers.
```
cb-ekuersch, line 5: we shouldn't need a version plan right? Its just infra/scripts that are changing in this PR

packages/illustrations/package.json
adrienzheng-cb: It's a private workspace lib at `libs/figma-api` (`private: true`), declared as a devDependency exactly like `packages/icons` does. Yarn resolves `workspace:^` on publish, so it ships as `^0.0.0` — already true of the published `@coinbase/cds-icons`, and harmless since consumers never install devDependencies. Let me know if you meant something else by "check this".

packages/illustrations/project.json
adrienzheng-cb: Yes — `@nx/jest:jest` is used by `icons`, `common`, `migrator`, `eslint-plugin-cds`, `figma-api`, and four others. Only web/mobile/android/ios use `nx:run-commands`, since they have their own runners. This target is identical to the one in `packages/icons`.

.nx/version-plans/illustrations-sync-tooling.md
adrienzheng-cb: Fair point that scripts shouldn't drive a release, and the trigger here is `packages/illustrations/package.json` rather than the scripts — the bump heuristic matches the path without reading the file, so a build-script devDependency looks publishable.

I looked at exempting devDependencies-only changes, but requiring a plan doesn't actually cut a release: plans accumulate until someone runs `yarn release`, and nx keeps the largest bump among them, so this patch gets absorbed into whatever illustrations ships next. Keeping it costs a changelog line and leaves the check strict enough to catch a build-tool bump that quietly changes published output — so I left the gate alone and kept the plan.

cb-ekuersch (approved):

PR #875, opened 2026-09-04
fix(utils): keep capitalize from lowercasing the rest of the string

pucedoteth:
## What

`capitalize` is declared as returning `Capitalize<T>`, which uppercases only the first character and leaves the remainder untouched. The implementation lowercases everything after it:

```ts
export const capitalize = <T extends string>(str: T): Capitalize<T> => {
  return (str.charAt(0).toUpperCase() + str.toLowerCase().slice(1)) as Capitalize<T>;
};
```

So the value contradicts its own type:

| call | declared type | actual value |
|---|---|---|
| `capitalize('USD')` | `'USD'` | `'Usd'` |
| `capitalize('ETH')` | `'ETH'` | `'Eth'` |
| `capitalize('helloWorld')` | `'HelloWorld'` | `'Helloworld'` |
| `capitalize('a11yLabel')` | `'A11yLabel'` | `'A11ylabel'` |

Because the declared type is a literal, TypeScript accepts the result wherever the original literal is required, while the runtime value differs. The `as Capitalize<T>` assertion is what lets that through.

## Why this direction

`capitalize` could equally have been meant as lodash-style ("first upper, rest lower"), in which case the type would be the thing to change. Two reasons for fixing the value instead:

1. The return type is explicit and matches a TypeScript built-in with defined semantics, and it is the part consumers can actually rely on at compile time.
2. It is behaviour-preserving for the pairing that exists in this module. `wordCase` is `decamelize(str, { separator: ' ' })`, whose output is already lowercase, so `capitalize(wordCase('helloWorld'))` is `'Hello world'` before and after.

The output only changes for inputs carrying uppercase characters past the first position, which are exactly the inputs whose current result violates `Capitalize<T>`.

If you would rather keep the lowercasing and relax the type instead, that is a one-line change in the other direction and I am happy to switch.

## Tests

`packages/utils` had no test files. Added `packages/utils/src/string.test.ts` covering `capitalize` (including the empty string, an already-capitalized input, and the `wordCase` pairing), the case helpers, and the css var helpers.

## Verification, and what I could not run

I could not run `yarn nx run utils:test` here. The repo is ~1&nbsp;GB and I worked from a partial clone, so the working tree is not complete enough to install and run the monorepo toolchain. Flagging that rather than implying the suite passed.

What I did do: executed every assertion in the new test file against the real `humps` build, since `camelCase`/`kebabCase`/`snakeCase`/`wordCase` delegate to it. All 13 hold, including the `capitalize(wordCase(...))` case. The `capitalize` change itself is one expression and is covered by the table above.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ None for this change
</td>
<td></td>
</tr>
</table>

pucedoteth:
Sorry for the noise here. I force-pushed this branch from a shallow clone, which produced a commit with no parent, so the diff rendered as the entire repository and the PR could no longer be reopened.

The branch is fixed and the change is now in #876: same two files, correct parent, signed. Please ignore this one.

PR #886, opened 2026-09-16
feat(mobile): add boxShadow support to the theme and fix elevation shadows

cb-ekuersch:
## What changed? Why?

Adds React Native's cross-platform `boxShadow` view style as an alternative form for the mobile theme's `shadow` tokens, fixes the duplicate Android shadow it exposed, lightens the `elevation1` shadow color on both platforms, and unbreaks the iOS simulator launch on Xcode 27 so the device check below is actually runnable.

`boxShadow` is a newer API based on web's spec by the same name. It is fully cross platform but requires the new Fiber RN acrhitecture.

## CDS Elevation levels (iOS left/Android right)

| master branch        | this branch w/o `boxShadow`        | this branch w/ `boxShadow` |
| -------------- | -------------- | ---- |
| shadowProp-master | shadowProps-branch | boxShadow-branch |

**1. `boxShadow` token form (`feat`, minor)**

`Shadow` becomes a never-padded discriminated union — a token declares either the legacy iOS-only `shadow*` fields or a `boxShadow` array, never both. Each arm declares the other arm's keys as `?: never`, so reads like `theme.shadow.elevation1.shadowRadius` keep resolving to `number | undefined` exactly as today, and every existing theme still satisfies `ThemeConfig` unchanged. Only the array form is accepted; RN parses the CSS string shorthand in JS and silently drops *every* shadow in a token on one bad unit.

`ChartText` converts either arm through the new `getSkiaShadowParams`, halving `blurRadius` to match iOS's `shadowRadius = blurRadius / 2` so charts and views agree, and warns in `__DEV__` when Skia's `<Shadow>` image filter can't express a `spreadDistance` or additional entries.

Built-in theme *values* are unchanged, so this is visually inert on its own.

**2. Duplicate Android shadow + `elevation` pass-through (`fix`)** — `Box` no longer writes a native Android `elevation` alongside a `boxShadow` token, and `elevation` is gone from `StyleProps`. See root cause below.

**3. Lighter `elevation1` shadow color (`fix`, patch)** — `elevation1` was shipping the same pure black as `elevation2` across every built-in theme on both platforms. Mobile `shadowColor` goes `#000000` → `#5B616E`; web goes `rgba(0, 0, 0, 0.12)` → `rgba(91, 97, 110, 0.12)` (the same color). `elevation2` is untouched. Native iOS/Android theme parity is tracked in [CDS-2650](https://linear.app/coinbase/issue/CDS-2650/update-ios-and-android-theme-elevation1-shadow-color-to-5b616e).

**4. iOS simulator launch on Xcode 27 (`fix`, tooling)** — Xcode 27 deleted Simulator.app and replaced it with Device Hub, breaking `open -a Simulator` in both the expo-app builder and the ios-gallery launch script. Both call sites now try Device Hub first and keep Simulator.app as a fallback for Xcode 26. The `@expo/cli` patch teaches the Expo CLI's own launcher the same thing so the `i` shortcut works; Expo SDK 57 fixes it upstream, drop that half of the patch on upgrade.

**Duplicate Android shadow.** `getElevationStyles` wrote a native `elevation: 2|8` unconditionally *and* spread the token. Legacy `shadow*` props are iOS-only, so Android previously saw exactly one shadow (the native one) — the bug was latent. A `boxShadow` token renders on Android by itself, so the native elevation drew a *second*, offset shadow on top of it.

The native elevation was originally kept on the theory that it drives draw order. It does not — z-order in react-native follows document order and `zIndex`. So `getElevationStyles` now emits the native elevation only for legacy tokens (which need it to paint anything on Android), gated on `isBoxShadowToken`. Legacy behavior is byte-identical until the deprecated props are removed in v11.

**`elevation` pass-through.** `elevation` was declared on `StyleProps` but registered in neither `themedStyleProps` nor `stylePropAliases`, so `getStyles` emitted the raw CDS level (`1` or `2`) as a native Android elevation. `Box` was unaffected in practice — `getElevationStyles` is appended later and wins — but `<Text elevation={1}>` emitted a bare Android `elevation: 1` and no shadow on either platform. Removing `elevation` from `StyleProps` closes it at the single choke point; `Text` keeps the prop as a deprecated, swallowed no-op for type compatibility.

This bug is mobile-only. On web, `elevation` is a registered *themed* style prop that resolves to `box-shadow: var(--shadow-elevationN)` + `background-color: var(--color-bgElevationN)`, with no raw passthrough and no second shadow mechanism.

**Xcode 27 simulator launch.** Device Hub needs a different *invocation*, not just a different app name: it ignores the legacy `-CurrentDeviceUDID` argument, so its `devices://` URL scheme is the only way to focus the device that was just booted. We probe rather than sniff the Xcode version, because a half-upgraded machine can keep a stale LaunchServices entry for the old Simulator.app path — making the version a less reliable signal than what actually resolves. Nothing claims `devices://` before Xcode 27, so `open` exits non-zero there and we fall through. Both branches are best-effort: booting and installing go through `simctl` and readiness is polled with `simctl bootstatus` rather than by watching for a window, so a failure to focus the UI never blocks the build.

### How has it been tested?

- [x] Unit tests
- [ ] Interaction tests
- [ ] Pseudo State tests
- [ ] Manual - Web

### Testing instructions

The existing `Box.test.tsx` shadow assertions pass **unmodified** apart from the intended `elevation1` color change — that's the backward-compatibility guarantee. New coverage in `packages/mobile/src/styles/__tests__/shadow.test.ts` (both arms of `isBoxShadowToken`, `getSkiaShadowParams` including the blur halving and the `undefined` case) and in `Box.test.tsx` (a `boxShadow`-authored theme emits `boxShadow` and **no** native `elevation`; legacy tokens still emit `elevation: 2|8`).

```
yarn nx run mobile:test --testNamePattern="Box|shadow|ChartText"
yarn nx run mobile:typecheck
```

Device check: `apps/expo-app` → Playground → **Box Shadow**. The new screen renders the full `BoxShadowValue` matrix (offset, blur, spread, color, multiple layers, inset) plus elevated CDS surfaces. Compare against `master` on an Android device to confirm the doubled shadow is gone.

Simulator launch (Xcode 27 should focus Device Hub, Xcode 26 should still open Simulator.app via the fallback):

```
yarn nx run expo-app:ios
yarn nx run ios-gallery:launch
```

Then press `i` in the Expo CLI and confirm it focuses the simulator too — that path is the `@expo/cli` patch.

> Note: `mobile:typecheck` currently fails on an unrelated dependency (`illustrations:typecheck` has 8 pre-existing errors under `scripts/sync-illustrations/`). `npx tsc -p packages/mobile/tsconfig.json --noEmit` is clean.

N/A — no files under `packages/illustrations/**` or `packages/icons/**` changed.

## Change management

type=routine 
risk=low 
impact=sev5 

automerge=false


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>1</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
<tr>
<td>
✅ See below
</td>
<td></td>
</tr>
</table>

<h2>
✅
<code>CODEOWNERS</code>
</h2>
<table>
<tr>
<th>Code Owner</th>
<th>Status</th>
<th>Calculation</th>
</tr>
<tr>

<td>ui-systems-eng-team</td>

<td>
✅
<code>1/1</code>
</td>


<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Sum</strong></td>
<td>
<strong>
0
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

packages/mobile/src/styles/shadow.ts
```diff
@@ -0,0 +1,76 @@
...
+/** Coerces react-native's `number | string` dimension values to a number, treating junk as zero. */
+const toNumber = (value: unknown): number => {
+  const parsed = Number(value);
+  return Number.isFinite(parsed) ? parsed : 0;
+};
+
+/**
+ * Resolves either shadow token form into the parameters Skia's `<Shadow>` image filter needs.
+ *
+ * Skia's `<Shadow>` can express a single shadow with no spread, so `boxShadow` tokens are reduced
```
adrienzheng-cb, line 45: nit: would it be better if we support adding multiple skia <Shadow /> if the boxShadow array includes more than one item

packages/mobile/src/visualizations/chart/text/ChartText.tsx
```diff
@@ -474,7 +475,16 @@ export const ChartText = memo<ChartTextProps>(
...
-    const elevationShadow = elevated ? theme.shadow.elevation1 : undefined;
+    const elevationShadow = useMemo(() => {
+      const shadow = elevated ? getSkiaShadowParams(theme.shadow.elevation1) : undefined;
+      if (!shadow) return undefined;
+
+      // getColorWithOpacity replaces the color's alpha rather than composing with it, so it is only
+      // correct for legacy tokens, where the opacity is authored as a separate value.
+      return shadow.opacity === undefined
+        ? shadow
+        : { ...shadow, color: getColorWithOpacity(shadow.color, shadow.opacity) };
```
adrienzheng-cb, line 486: should we just call getColorWithOpacity inside getSkiaShadowParams

adrienzheng-cb (dismissed):

packages/mobile/src/styles/styleProps.ts
```diff
@@ -64,7 +63,6 @@ export type StyleProps = {
   marginBottom?: NegativeSpace;
   marginStart?: NegativeSpace;
   marginEnd?: NegativeSpace;
-  elevation?: ElevationLevels;
```
hcopp, line 67: @cb-ekuersch why are we removing this?

hcopp: Specifically here, I do understand the overall shift going on for mobile

I worry this could potentially break some implementations right? Given it is a prop that is exported

packages/mcp-server/CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+## 9.28.0 (2026-09-18)
+
+### 🚀 Features
+
+- Support React Native boxShadow property on mobile theme and deprecates individual shadow properties - no default theme changes in this release
+- Feat: mobile shadow tokens can now be authored with React Native's cross-platform `boxShadow` view style instead of the iOS-only `shadowColor`/`shadowOffset`/`shadowOpacity`/`shadowRadius` props.
```
hcopp: small nit/opinion: feels very big for a changelog, especially for MCP server this doesn't feel applicable. Seems like we should say "This is an artificial version bump with no new change."

And then at most for mobile do a couple of lines and then link to docs if we need to express it further

packages/web/CHANGELOG.md
```diff
@@ -1,3 +1,32 @@
+## 9.28.0 (2026-09-18)
+
+### 🚀 Features
+
+- Support React Native boxShadow property on mobile theme and deprecates individual shadow properties - no default theme changes in this release
+- Feat: mobile shadow tokens can now be authored with React Native's cross-platform `boxShadow` view style instead of the iOS-only `shadowColor`/`shadowOffset`/`shadowOpacity`/`shadowRadius` props.
```
hcopp: Yeah I'm thinking there is a bug with our changelog generation, definitely shouldn't talk about mobile changes on web.

adrienzheng-cb (approved):

